PR #293, opened 2014-11-17
hmc_tm: add real reproducible random numbers mode

kostrzewa:
this will keep the RNG state consistent across executions of different numbers of trajectories (say you have a serial run doing 1000 trajectories at a time in 10 goes and a parallel run doing 10000 trajectories in one run, with this mode, the two runs will still have exactly the same random numbers because the RNG state is saved to file between executions of hmc_tm)

This works only when using reproducible_random_numbers and when the random numbers are requested only through functions supporting this mode.

### DIFF
--- a/default_input_values.h
+++ b/default_input_values.h
@@ -54,8 +54,9 @@
 #define _default_g_beta 6.0
 #define _default_g_N_s 20
 #define _default_g_dflgcr_flag 0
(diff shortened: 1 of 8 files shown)


kostrzewa:
Your proposal is still missing, other than that, I've tested this at the time.

PR #297, opened 2015-01-12
read_input: make rmQuotes static inline void to satisfy Cray compiler

kostrzewa:
The Cray compiler complains about not being able to find the rmQuotes function. Declaring it ```static inline void``` seems to fix the problem.

### DIFF
--- a/read_input.l
+++ b/read_input.l
@@ -63,7 +63,7 @@ EQL {SPC}*={SPC}*
 #include"phmc.h"
 #include<io/params.h>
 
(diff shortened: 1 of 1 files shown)


kostrzewa:
In fact, I believe this is correct C99 behaviour. The alternative would be to have 

```
inline void rmQuotes(...) 
{
}
```

in some header file and to declare it in ```read_input.l``` using ```external inline void rmQuotes(...)```

PR #299, opened 2015-02-18
non even/odd clover det monomial and non even/odd clover operator

urbach:
This implements
  * a not even/odd preconditioned clover operator (from Abdou)
  * the CG solver for this non even/odd clover operator
  * a non even/odd clover det monomial


urbach:
I think this is tested now for the case of 
  * no SSE
  * no MPI

urbach:
to wait for pull request #306

PR #300, opened 2015-02-19
Implementation of the gradient flow

kostrzewa:
Currently only the Wilson flow is implemented (the extension should be trivial I believe to also allow Iwasaki, Symanzik and DBW2 flows) and no parameters are provided (e.g., we always integrate to t=10)


kostrzewa:
One could also choose to do one more step for each iteration of the while loop and output a higher density of integration steps (currently only t=0.01,0.03...9.99 are printed)

urbach:
hmm, 'matrix_utils.c' seems to be missing?

kostrzewa:
oops

kostrzewa:
Sorry about that, since HLRN doesn't support connecting to github directly I simply copied my local source code there which included matrix_utils.[c,h]

urbach:
No problem. I'll check again. Do you have this compiled on juqueen already, such that I can use it directly?

kostrzewa:
No I'm afraid I haven't prepared a merge of this with InterleavedNDTwistedClover

kostrzewa:
Oh, and when you do that, please don't forget to merge in the latest stuff from master as well, because otherwise tracelog will be incorrect.

urbach:
on juqueen the subdirectory `meas` was not created for some reason in the build directory. I don't understand yet, why...

urbach:
otherwise `InterleavedNDTwistedClover` meged with `gradient_flow` compiles on juqueen as well.

kostrzewa:
Hmm, did you build it in a new build directory? If yes then maybe configure was not regenerated.

urbach:
yes, new build directory

kostrzewa:
That might be it then, I've had that happen to me every time we introduced a new subdirectory :)

urbach:
yes, I remember this as well... So, I'm merging this here...

kostrzewa:
Great, thanks for checking and fixing. At some point in the future I'm sure we can add the other gradient flows and add a parameter for the maximum integration point.

PR #302, opened 2015-06-08
sse.h was not included when configured for SSEX usage

urbach:
hmm, somehow I never tested this with SSE ...

We could also include `global.h` instead, but this would create more dependencies.

### DIFF
--- a/matrix_utils.c
+++ b/matrix_utils.c
@@ -25,6 +25,9 @@
 #include <math.h>
 #include <complex.h> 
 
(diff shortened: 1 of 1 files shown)


kostrzewa:
Oops! Although I have to say that having ```sse.h``` to be included everywhere is a bit strange, but I guess it's the same with all the other stuff that we have.

urbach:
well, usually it comes with `global.h`...

PR #305, opened 2015-11-23
Degenerate mixed CG for operators

kostrzewa:
This is the first in a series of pull-requests through which I would like to merge in Florian's mixed CG. I would like to avoid having one massive pull-request.This contains the MCG for TMWILSON and CLOVER operators. It is currently broken for ```--enable-sse2/3``` because the inline assembly must be rewritten. Usage of the Intel compiler is thus strongly recommended. It should work unmodified also using the Cray compiler, which is likely to provide similar performance.

It works on BG/Q, but does not include overlapping communication and computation. An attempt is made at exploiting orphaned OMP directives for the single precision operators and this seems to work well, but the overhead reduction is currently unknown.

It has been subjected to very basic tests (checking the real residual) and no problems have been observed.

The orphaning approach might be the basis of our overall deployment strategy for orphaned directives, but tests and benchmarks on various architectures are required before I'm willing to push this type of modification also into the double implementation.

Eventually I would like to merge Florian's GPU work because it would be very beneficial for us if we were able to do the whole simulation on Hybrid machines with a few cards on moderate lattice volumes, perhaps even up to L=48.


kostrzewa:
@urbach: I guess before this is pulled in I should either write the SSE inline (which I doubt I will have time for) or simply use some preprocessor constants to undef and redef SSE in the single precision operators.

kostrzewa:
Okay, so this is ready to be merged now. Monomials will follow soon!

kostrzewa:
Forgot to make the modification which led to segfaults on BG/Q. Rectified.

urbach:
can you also provide sample input files (which you probably have anyhow)!?

kostrzewa:
good idea, will do

kostrzewa:
There's the sample input file.

PR #306, opened 2016-01-15
Merge deflation

urbach:
okay, this is a long commit :( 

It provides a running inexact deflation solver for TM and TM+CLOVER (single flavour). It is tested serially and MPI parallel, including also OpenMP and hybrid! 

This commit also includes a non-even/odd clover operator implementation.

A point for discussion is where to call the deflation subspace generation and sw_invert etc. I don't think that invert.c is the best place for this.


kostrzewa:
I'm sorry the review is taking so long, I'll look at is ASAP.

solver/gcr4complex_body.c
```diff
@@ -0,0 +1,321 @@
...
+        break;
+      }
+    }
+    /* prepare for restart */
+    _PSWITCH(c)[k] /= _PSWITCH(b)[k];
+    _PSWITCH(lassign_add_mul)(P, _PSWITCH(xi)[k], _PSWITCH(c)[k], N);
+    for(l = k-1; l >= 0; --l) {
+      for(i = l+1; i <= k; ++i) {
+        ctmp  = _PSWITCH(a)[l][i] * _PSWITCH(c)[i];
+        _PSWITCH(c)[l] -= ctmp;
```
urbach: @kostrzewa do you think we should do some of the linalg here in double also in the 32 Bit version?

solver/bicg_complex.c
```diff
@@ -0,0 +1,128 @@
...
+#include "solver_field.h"
+#include "bicg_complex.h"
+
+/* P inout (guess for the solving spinor)
+   Q input
+ */
+int bicg_complex(spinor * const P,spinor * const Q, const int max_iter, 
+		double eps_sq, const int rel_prec, 
+		const int N, matrix_mult f, matrix_mult fdagg){
+	double err, squarenorm;
```
kostrzewa: Were the tabs intended?

solver/gcr4complex_body.c
kostrzewa: I don't know about GCR, but in the CG solver, doing any of the sub-steps in high precision does not seem to help much. What's important, as far as I understand, is keeping the solution in high precision (either directly or incrementally through reliable updates). In the latter case then one of course needs a zero starting vector for the iterated solutions in low precision.

urbach: okay, here we want to solve to a loose precision only anyhow, 10^-2 or so...

solver/bicg_complex.c
urbach: eh, no, I'll fix that (again). It sneaks in over again...

urbach:
I think the 32bit improvements are basically there. My tests are positive so far, all seems to reproduce the results in full precision. It probably still needs some optimisation and testing.

urbach:
we have to find out whether 'lgcr' or 'lfgmres' is better

urbach:
Its not totally clear to me how to incorporate the deflated solver into the HMC: the problem is the update of the deflation subspace and how to trigger it in the current interface.

urbach:
Its not totally clear to me how to incorporate the deflated solver into the HMC: the problem is the update of the deflation subspace and how to trigger it in the current interface.

kostrzewa:
In principle there is already a structure for keeping a history: the chronological solver. Can we share the deflation subspace for several monomials? If yes, then the structure would be rather simple I think. But I realise that this will take some design work...

urbach:
the chronological solver keeps the history of the solution spinors. I was more after the update of the deflation subspace. I guess we have to update it every n-th delta tau step. On which time scale? And how does the monomial know about it? 

deflation should be usable for several monomials.

kostrzewa:
> the chronological solver keeps the history of the solution spinors. I was more after the update of the deflation subspace. I guess we have to update it every n-th delta tau step. On which time scale? And how does the monomial know about it? 

That was sort of my point: IF the same deflation subspace can be used by several monomials (with different rho or 2kappamu respectively), as you say should be the case, then the keeping track of the number of steps (which should be the number of steps on timescale 0, because the subspace depends on the gauge field) can be taken care of by an extension of the chronological mechanism. When the deflated solver is selected, it gets the subspace from the chronological mechanism and passes it to it respectively. If the subspace is out of date or does not exist, it is the chronological solver which must contain the logic of regenerating it. Of course, this requires a substantial extension of the chronological interface or the addition of something similar. I would be more than happy to contribute significantly to this because I find it highly interesting from an engineering p.o.v., but I don't know how I can fit it in with all the other projects...

kostrzewa:
I would also advocate that the code be general enough to support also the non-degenerate doublet and therefore also a second subspace, so that it can be used for the most expensive shifts, as done in OpenQCD.

urbach:
I'm not sure I agree that this job should be taken by the chronological solver. This would, among other things, mean we cannot use deflation without chronological solver.

In OpenQCD things are easier, because the subspace is the same for all flavours. For us, doing it also for s/c will need significant additions. Its currently not foreseen...

urbach:
@kostrzewa maybe this could be merged at some point...?

kostrzewa:
@urbach Sure, it compiles just fine even on BG/Q. There are minor things such as commented out code sections (in block.c, for example) and comments like "NEEDS TO BE REWRITTEN" which are a little confusing. I don't know if these should not be cleaned up before pulling this in.

urbach:
okay, let me have a look...

> @urbach Sure, it compiles just fine even on BG/Q. There are minor things such as commented out code sections (in block.c, for example) and comments like "NEEDS TO BE REWRITTEN" which are a little confusing. I don't know if these should not be cleaned up before pulling this in.
> 
> ---
> https://github.com/etmc/tmLQCD/pull/306#issuecomment-219956812
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

block.c
```diff
@@ -134,20 +148,21 @@ int init_blocks(const int nt, const int nx, const int ny, const int nz) {
   if((void*)(u = (su3*)calloc(1+8*VOLUME, sizeof(su3))) == NULL) {
     CALLOC_ERROR_CRASH;
   }
+  if((void*)(u_32 = (su3_32*)calloc(1+8*VOLUME, sizeof(su3_32))) == NULL) {
```
kostrzewa: shouldn't this also be aligned memory?

invert_eo.c
```diff
@@ -66,21 +66,21 @@ static void cgmms_write_props(spinor ** const P, double const * const extra_mass
...
 #endif
 
 int invert_eo(spinor * const Even_new, spinor * const Odd_new, 
-        spinor * const Even, spinor * const Odd,
-        const double precision, const int max_iter,
-        const int solver_flag, const int rel_prec,
-        const int sub_evs_flag, const int even_odd_flag,
-        const int no_extra_masses, double * const extra_masses, solver_params_t solver_params, const int id,
-        const ExternalInverter inverter, const SloppyPrecision sloppy, const CompressionType compression )  {
+	      spinor * const Even, spinor * const Odd,
```
kostrzewa: tabs

linalg/scalar_prod_body.c
```diff
@@ -0,0 +1,130 @@
...
+
+  ks = 0.0;
+  kc = 0.0;
+
+#if (defined BGL && defined XLC)
+  __alignx(16, S);
+  __alignx(16, R);
+#endif
+
+#ifdef OMP2
```
kostrzewa: Is ```OMP2``` here intended?

block.c
urbach: its aligned later on, isn't it? But I can change to aligned_malloc directly...

little_D_body.c
```diff
@@ -0,0 +1,131 @@
...
+    g_mu = 0.;
+    compute_little_D(1);
+    g_mu = musave;
+    dfl_subspace_updated = 0;
+  }
+  _PSWITCH(little_D)(tmp, w);
+  _PSWITCH(little_D)(v, tmp);
+  aligned_free(tmp);
+  _PSWITCH(lassign_add_mul)(v, w, g_mu*g_mu + g_mu2*g_mu2, nb_blocks*g_N_s);
+  //memcpy(v, w, nb_blocks * g_N_s*sizeof(_C_TYPE));
```
kostrzewa: stale line?

monomial/cloverdet_monomial.c
```diff
@@ -89,39 +95,46 @@ void cloverdet_derivative(const int id, hamiltonian_field_t * const hf) {
   chrono_guess(mnl->w_fields[1], mnl->pf, mnl->csg_field, mnl->csg_index_array,
 	       mnl->csg_N, mnl->csg_n, VOLUME/2, mnl->Qsq);
   mnl->iter1 += solve_degenerate(mnl->w_fields[1], mnl->pf, mnl->solver_params, mnl->maxiter, mnl->forceprec, 
-		       g_relative_precision_flag, VOLUME/2, mnl->Qsq, mnl->solver);
+				 g_relative_precision_flag, VOLUME/2, mnl->Qsq, mnl->solver);
```
kostrzewa: tabs have appeared

solver/mcr4complex.c
```diff
@@ -0,0 +1,216 @@
...
+  double norm_sq, norm,err;
+  _Complex double ctmp;
+  _Complex double alpha,beta;
+  _Complex double one = 1.0;
+
+
+  double atime, etime;
+  init_lmcr(m, lda);
+
+  //    // XDU 2011/10/03 DEBUG
```
kostrzewa: these are not necessary anymore, I guess

xchange/little_field_gather_body.c
```diff
@@ -0,0 +1,433 @@
...
+    ib2=0;
+    for(int bt = 0; bt < nblks_t; bt++) {
+      for(int bx = 0; bx < nblks_x; bx++) {
+        for(int by = 0; by < nblks_y; by++) {
+          for(int bz = 0; bz < nblks_z; bz++) {
+            if ((bt+bx+by+bz)%2==eo) {
+              ib2 = index_block_eo[block_index(bt, bx, by, bz)] * g_N_s;
+              
+              switch(pm){ 
+              case 0: /* Direction +t */
```
kostrzewa: use enum values here?

xchange/little_field_gather.c
```diff
@@ -0,0 +1,80 @@
...
+#ifdef MPI
+# include <mpi.h>
+#endif
+#include "global.h"
+#include <complex.h>
+#include "block.h"
+#include "little_field_gather.h"
+
+enum{
+  NONE = 0,
```
kostrzewa: NONE = 0 seems a bit strange, is this correct?

xchange/little_field_gather_body.c
```diff
@@ -0,0 +1,433 @@
...
+  /* We proceed like this for code simplicity, maybe will be optimized later */
+  
+  for(int pm = 0; pm < 8; pm++) {
+    for(int bt = 0; bt < nblks_t; bt++) {
+      for(int bx = 0; bx < nblks_x; bx++) {
+        for(int by = 0; by < nblks_y; by++) {
+          for(int bz = 0; bz < nblks_z; bz++) {
+            ib = block_index(bt, bx, by, bz) * g_N_s;
+            switch(pm){ 
+            case 0: /* Direction +t */
```
kostrzewa: use enum values here? but what about NONE=0?

kostrzewa:
I've given the whole changeset a rough read-through and can't see anything obviously wrong.

Added some comments on the block MPI routines and the ```Direction``` enum that has been defined (but not used). Adjusting this might change behaviour so this should be done carefully, if it is done.

block.c
kostrzewa: Sure, however you prefer. Using the latter makes the code more readable of course.

urbach:
hmm, the `Direction` enum is used in the MPI routines, but not in the `if` clauses. Looks incompatible...

urbach:
ah, in the MPI routines its only used as the tag, so it can be changed.

kostrzewa:
Okay, pulling it in. Thanks!

PR #307, opened 2016-02-08
monomial components for mixed cg

kostrzewa:
Components to use mixed solvers for all determinant and non-degenerate rational approximation type monomials. Note that in its current form it significantly increases memory usage because the single precision fields are always allocated.

### DIFF
--- a/default_input_values.h
+++ b/default_input_values.h
@@ -171,7 +171,7 @@
 
 /* default mixed precision solver values */
 #define _default_mixcg_innereps 1.0e-6
-#define _default_mixcg_maxinnersolverit 1500
+#define _default_mixcg_maxinnersolverit 5000
 
 #define _default_use_preconditioning 0
 
--- a/hmc_tm.c
+++ b/hmc_tm.c
@@ -164,7 +164,10 @@ int main(int argc,char *argv[]) {
 
   DUM_BI_MATRIX = DUM_BI_SOLVER+6;
   NO_OF_BISPINORFIELDS = DUM_BI_MATRIX+6;
-
+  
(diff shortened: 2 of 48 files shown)


urbach:
hmm, why does it have conflicts with master?

kostrzewa:
Oh, just because of the merge you did a few moments ago, not to worry. Let's keep it unmerged for now, I'll fix it up as I go.

kostrzewa:
So I have started a test run of the mixed mms for ND*RAT*, but I think I need to add a few safeguards to the solver to prevent incorrect residuals being accepted. (I don't yet know if the problem even occurs, but better safe than sorry...)

kostrzewa:
I'm afraid the speed-up factor is small, maybe I need to change the ordering of the operator so that rather than being grouped by function, the operator is grouped by data.

kostrzewa:
Using the mixed mms on a small test volume, the residuals for all shifts other than the guiding shift are of O(1e-10) or a bit higher. Conversely, with the non-mixed solver, these residuals are all significantly smaller than that of the guiding shift. I don't know if this explains the dH, but it seems to me that the mixed solver in its present form is not able to iterate the higher shifts sufficiently.

kostrzewa:
On the production volume (L=32) some of the residuals are O(1) or larger! I've checked the operator 
for the square norm of the operator applied to some spinor, the resulting vector has square norm 
##### ||M_ndpsi X|| = ||M_ndpsi32 X_32|| ~ 3e7
while the square norm of the difference comes out as
##### ||M_ndpsi X - M_ndpsi32 X_32|| ~ 1e-8,
which I think matches the expectation from round-off, no?

PR #309, opened 2016-02-25
Additions for lib_wrapper and the QUDA interface

kostrzewa:
* Added EnableSubprocess option which allows the passing of an external
  unique ID to tmLQCD_invert_init such that an external MPI program
  can link against the SERIAL version of tmLQCD and the MPI id of the
  calling program is passed along. For now, this is used only
  for selecting the QUDA device, but could also be used for I/O
  purposes or other special usecases. Introduces the global variable
  g_external_id.

* Added EnableLowmem option which initialises tmLQCD with only the
  gauge field memory (no copy) and the essential geometry fields.
  Useful for running with the QUDA interface on low memory machines
  ( < 64 GB RAM ). Lots of warnings are printed to warn the user
  about what they are doing.

* Switched to QUDA_NORMERR_*_SOLVE parameter in QUDA interface such
  that the resduals between QUDA and tmLQCD match exactly. This
  now solves (A A^dag) y = b -> x = A^dag y rather than
  (A^dag A) x = A^dag b, just like tmLQCD.

* Added quda_invert_direct function such that tmLQCD can be used
  as a thin wrapper around the QUDA library with input and output
  in the lexicographical tmLQCD ordering (TXYZsc) with pointers
  to double. The input file must still be loaded and an operator
  defined appropriately.


kostrzewa:
Please do not merge yet, I will modify this somewhat so that it can also be used with MPI enabled.

urbach:
hmm, has conflicts now, of course. Can this be merged now at some point?

kostrzewa:
sure, will resolve soon, maybe today

kostrzewa:
@urbach I've resolved the conflicts as far as I can tell. Because I need to do some stuff with peram_gen, it would be nice to merge this in soon. Cheers.

kostrzewa:
I have to say, I don't like the lowmem and subprocess solutions, but I don't know how else to achieve the reduction in memory usage / ability to use single GPUs for highest efficiency differently...

kostrzewa:
There seem to be some compilation problems...

urbach:
okay, looks okay to me. Someone should test this on GPUs, though. I'll pull it in, because I think it does not affect anything else.

kostrzewa:
It affects anyone using lib_wrapper because the signature of tmLQCD_invert_init has changed, other than that, you are correct.

urbach:
We could also print only the hash. ..

On 7 July 2016 16:40:52 CEST, Bartosz Kostrzewa <notifications@github.com> wrote:
>It affects anyone using lib_wrapper because the signature of
>tmLQCD_invert_init has changed, other than that, you are correct.
>
>---
>You are receiving this because you modified the open/close state.
>https://github.com/etmc/tmLQCD/pull/309#issuecomment-231098260

kostrzewa:
> We could also print only the hash. ..

I was referring to the quda_interface additions that I did (+ lowmem and subprocess mode). The latter adds the "external ID" as one of the arguments of tmLQCD_invert_init

PR #318, opened 2016-05-09
Reweighting

urbach:
stripped down cloverdetratio monomial for reweighting


kostrzewa:
Looks fine to me.

kostrzewa:
Actually, I think the input file reader does not expose kappa2 for the new monomial

urbach:
> Actually, I think the input file reader does not expose kappa2 for the new monomial

It does, but there is another problem I don't understand...

PR #320, opened 2016-05-10
Reweighting

urbach:
some more improvements

### DIFF
--- a/reweighting_factor.c
+++ b/reweighting_factor.c
@@ -57,20 +57,20 @@ void reweighting_factor(const int N, const int nstore) {
       double c_sw = mnl->c_sw;
       if(c_sw < 0.) c_sw = 0.;
 
-      sw_term( (const su3**) hf.gaugefield, mnl->kappa2, c_sw); 
+      sw_term( (const su3**) hf.gaugefield, mnl->kappa, c_sw); 
       if(mnl->type != NDDETRATIO) {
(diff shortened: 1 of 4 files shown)


kostrzewa:
In order to study the error scaling with the number of reweighting samples without having to rerun the full computation each time, I think it would be useful to extend the output further and either provide seperate files for each sample or a sample index in the output file. This of course also means extending the ```sum[j]``` and ```sum_sq[j]``` arrays with another index.

reweighting_factor.c
```diff
@@ -83,13 +83,15 @@ void reweighting_factor(const int N, const int nstore) {
   }
   
   if(g_proc_id == 0) {
-    ofs = fopen("reweighting_factor.data", "a");
-    fprintf(ofs, "%d ", nstore);
+    char filename[50];
+    sprintf(filename, "reweighting_factor.data.%.5d", nstore);
+    ofs = fopen(filename, "w");
```
kostrzewa: file open check?

urbach:
ah, the trlog part is still missing, isn't it?

kostrzewa:
True, with different kappas is doesn't cancel...

urbach:
@kostrzewa could you crosscheck whether I got the signs correct?

kostrzewa:
Thanks, as discussed "offline", this looks fine now.

Issue #314, opened 2016-03-21
MPI and OMP should be renamed


kostrzewa:
```MPI``` and ```OMP``` should be renamed, probably to ```TM_USE_MPI``` and ```TM_USE_OMP```

urbach:
see #322

urbach:
see #323

PR #322, opened 2016-05-19
Replace MPI macro definition

urbach:
partly solves #314


urbach:
what I did was to replace

    "ifdef[blank]* MPI" -> "ifdef TM_USE_MPI"
    "ifndef[blank]* MPI" -> "ifndef TM_USE_MPI"
    "defined[blank]* MPI" -> "defined TM_USE_MPI"

and some manual replacements in the GPU directory. And of course corresponding changes to configure.in and config.h.in

kostrzewa:
Conflicts from #323

urbach:
fixed!

--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
Looks okay to me, checked ```grep MPI *.c *.h */*.c */*.h | grep def | less```

kostrzewa:
I'll merge this now. Propagating the changes to the interleaved branches will be a little bit tricky because of the three loop halfspinor hopping matrix, which will likely have to be adjusted manually.

urbach:
True,  didn't think that far yet

On 22 May 2016 14:13:59 CEST, Bartosz Kostrzewa <notifications@github.com> wrote:
>I'll merge this now. Propagating the changes to the interleaved
>branches will be a little bit tricky because of the three loop
>halfspinor hopping matrix, which will likely have to be adjusted
>manually.
>
>---
>You are receiving this because you were assigned.
>https://github.com/etmc/tmLQCD/pull/322#issuecomment-220829188

urbach:
so, indeed, the file `halfspinor_body.c` is causing problems in the interleaving branch. But it seems to me this can be fixed by just doing the replacements of the macros again in this file? What do you think @kostrzewa

urbach:
see my branch testMerge which originates from the interleaved branch. Do you think `halfspinor_body.c` is okay like this?

kostrzewa:
@urbach 
> But it seems to me this can be fixed by just doing the replacements of the macros again in this file?

I agree, the "natural" way to resolve the merge conflict would be to choose the "ours" side of the merge for the two halfspinor bodies (64 and 32) and manually replace all mentions ```MPI``` and ```OMP```. 

However, ```halfspinor_body.c``` is fine (did a side-by-side diff, only differences are the macros), but ```halfspinor_body_32.c``` in your INDTC branch is missing the overlapping structure. When Florian re-implemted the 32 bit hopping matrix, he took a slightly different approach as well, which I preserved. (the main difference are some prefetches)

See here: https://github.com/kostrzewa/tmLQCD/blob/mixed_cg_merge_hmc_INDTC/operator/halfspinor_body_32.c

urbach:
so, which version of `halfspinor_body_32.c` I should start from then?

kostrzewa:
I think if you ```checkout``` the one from my mixed_cg_merge_hmc_INDTC branch into your testMerge branch and do the replacements, that should be fine. That's the version we have been using all this time.

urbach:
okay, done...

kostrzewa:
Okay, thanks a lot!

urbach:
so, I'll merge this back to my interleaving branch, I guess...

urbach:
done

PR #324, opened 2016-05-19
removed last instances of DUM_SOLVER

urbach:


urbach:
what I did was to replace

    "ifdef[blank]* MPI" -> "ifdef TM_USE_MPI"
    "ifndef[blank]* MPI" -> "ifndef TM_USE_MPI"
    "defined[blank]* MPI" -> "defined TM_USE_MPI"

and some manual replacements in the GPU directory

kostrzewa:
Hmm, do you think we should do something about this in general? ```DUM_DERI``` seems even stranger than ```DUM_SOLVER```...

urbach:
In principle, yes, but DUM_SOLVER I had removed more or less already some time ago.

Maybe I'll give the buffer thingy a go at some point. 

On 27 May 2016 21:17:39 CEST, Bartosz Kostrzewa <notifications@github.com> wrote:
>Hmm, do you think we should do something about this in general?
>```DUM_DERI``` seems even stranger than ```DUM_SOLVER```... 
>
>---
>You are receiving this because you authored the thread.
>https://github.com/etmc/tmLQCD/pull/324#issuecomment-222231498

urbach:
I think this could be merged without affecting anything...

kostrzewa:
Looks fine to me, I'll merge this now.

PR #326, opened 2016-05-22
merge pion <x> branch

kostrzewa:


kostrzewa:
I've added support for differing propagator and source filenames. It seems to work fine now. I also fixed some apparent logic bugs in the process.

invert.c
```diff
@@ -367,7 +367,7 @@ int main(int argc, char *argv[])
     if (g_dflgcr_flag) 
       init_blocks(nblocks_t, nblocks_x, nblocks_y, nblocks_z);
     
-    if(SourceInfo.type == SRC_TYPE_VOL) {
+    if(SourceInfo.type == SRC_TYPE_VOL || SourceInfo.type == SRC_TYPE_PION_TS || SourceInfo.type == SRC_TYPE_GEN_PION_TS) {
```
kostrzewa: I've added this because it seems that otherwise for pion_ts and gen_pion_ts, the inverter would run twelve times unless Indices was set manually to 0. Not sure why that was not the case when Max and @urbach did the <x> inversions.

prepare_source.c
```diff
@@ -51,21 +52,17 @@ void prepare_source(const int nstore, const int isample, const int ix, const int
...
+  float u;
   SourceInfo.nstore = nstore;
   SourceInfo.sample = isample;
   SourceInfo.ix = ix;
 
   if(optr->type != DBTMWILSON && optr->type != DBCLOVER) {
     SourceInfo.no_flavours = 1;
     /* no volume sources */
-    if(source_type != 1) {
+    if(source_type == SRC_TYPE_POINT || source_type == SRC_TYPE_TS) {
```
kostrzewa: @urbach: It seems to me that this is appropriate for the logic, what do you think?

prepare_source.c
```diff
@@ -146,6 +143,74 @@ void prepare_source(const int nstore, const int isample, const int ix, const int
       }
       sprintf(source_filename, "%s.%.4d.%.5d.inverted", PropInfo.basename, nstore, isample);
     }
+    else if(source_type == SRC_TYPE_PION_TS) {
+      // Pion full time slice sources
+      if(SourceInfo.automaticTS) {
```
kostrzewa: @urbach: I find this a bit strange, why would it be appropriate to loop for ```SRC_TYPE_PION_TS```? It i sclear that ```SRC_TYPE_GEN_PION_TS``` requires the loop.

read_input.l
```diff
@@ -2449,9 +2462,9 @@ int read_input(char * conf_file){
   X3 = _default_X3;
   g_rgi_C1 = _default_g_rgi_C1;
   read_source_flag= _default_read_source_flag;
-  if(SourceInfo.basename == NULL) SourceInfo.basename = (char*)malloc(strlen(_default_source_filename)+10);
```
kostrzewa: @urbach: I think here malloc should always be called.

read_input.l
```diff
@@ -2084,13 +2094,16 @@ static inline void rmQuotes(char *str){
   if(myverbose!=0) printf("Don't read inversion source from file, but save the one generated\n");
 }
 <SOURCEFILE>{FILENAME} {
-  if(SourceInfo.basename == NULL) free(SourceInfo.basename);
```
kostrzewa: @urbach: Here free should only be called if the pointer is non-NULL, I guess. Not sure why it was the other way around before...

urbach: looks like a typo to me

read_input.l
urbach: why also in case of a re-read?

prepare_source.c
urbach: Hmm, I don't really understand the question

kostrzewa: To clarify my thinking: When using multiple samples, the source timeslice should change from sample to sample, no? The down prop is generated within the loop in ```op_invert```, so the same source (and timeslice) would be used regardless (prepare_source would not even be called).

kostrzewa: > Hmm, I don't really understand the question

What I mean is: why would it be appropriate to attempt to detect the source timeslice for ```SRC_TYPE_PION_TS``` from an existing propagator? I would expect that for this source type, the only appropriate course of action is drawing a random number and choosing a source timeslice, as done when ```found==0``` after the loop.

read_input.l
kostrzewa: Ah, I didn't consider this. Okay, I'll change it back!

kostrzewa: ```reread_input()``` does not set default values, it simply calls ```parse_config()``` which itself then skips a number of initialisations. I can also change this back because ```SourceInfo.basename``` and ```PropInfo.basename``` are set to ```NULL``` before ever being used.

kostrzewa:
I believe this is fine now.

prepare_source.c
urbach: yes, as discussed in person...

urbach:
okay, seems alright, I'm pulling this in now

kostrzewa:
Thanks!

PR #327, opened 2016-05-24
found overlooked instances of defined MPI

urbach:
we missed one instance of MPI, which was written as

    defined(MPI)

which is corrected now


kostrzewa:
Thanks!

PR #334, opened 2016-07-06
Source Seed

urbach:
this is making the random numbers used in the function `source_generation_pion_only` depending on `random_seed`. Note that once this is merged, we loose reproducability to previous computations using pion times slice sources.


kostrzewa:
For the sequential propagator, do I understand correctly that in principle the choice of time slice is not yet available, but the interface has been prepared accordingly?

kostrzewa:
For the rest, I'm in favour of merging this.

urbach:
okay, as discussed, go for it...

PR #335, opened 2016-07-06
make lib_wrapper write parameters and git_hash to stdout and to parameter file

urbach:


kostrzewa:
Looks good and compiles fine, merging.

PR #336, opened 2016-08-04
DO NOT MERGE first test of the ndrat monomial with a shift by shift solve

kostrzewa:
first test of the ndrat monomial with a shift by shift solve rgmixedcg

I'm sure I'm doing something wrong, but would this work in principle? See `solve_mms_nd`. Generalisations need to be made for the non-clover matrix, of course, and it probably requires some special treatment for the acceptance step. DO NOT MERGE :)

### DIFF
--- a/default_input_values.h
+++ b/default_input_values.h
@@ -50,6 +50,7 @@
 #define _default_g_mu1 0.0
 #define _default_g_mu2 0.0
 #define _default_g_mu3 0.0
+#define _default_g_shift 0.0
 #define _default_c_sw -1.0
 #define _default_g_beta 6.0
 #define _default_g_N_s 20
--- a/global.h
+++ b/global.h
@@ -194,7 +194,7 @@ EXTERN su3adj ** ddummy;
 
 EXTERN int count00,count01,count10,count11,count20,count21;
 EXTERN double g_kappa, g_c_sw, g_beta;
-EXTERN double g_mu, g_mu1, g_mu2, g_mu3;
+EXTERN double g_mu, g_mu1, g_mu2, g_mu3, g_shift;
 EXTERN double g_rgi_C0, g_rgi_C1;
 
 /* Parameters for non-degenrate case */
--- a/linalg/assign_add_mul_r_32.c
+++ b/linalg/assign_add_mul_r_32.c
@@ -37,13 +37,8 @@
 #include "su3.h"
 #include "assign_add_mul_r_32.h"
 
-
 #if (defined BGQ && defined XLC)
-void assign_add_mul_r_32(spinor32 * const R, spinor32 * const S, const float c, const int N) {
-#ifdef TM_USE_OMP
-#pragma omp parallel
-  {
-#endif
+void assign_add_mul_r_32_orphaned(spinor32 * const R, spinor32 * const S, const float c, const int N) {
   vector4double x0, x1, x2, x3, x4, x5, y0, y1, y2, y3, y4, y5;
   vector4double z0, z1, z2, z3, z4, z5, k;
   float *s, *r;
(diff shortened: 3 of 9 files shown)


urbach:
I don't quite understand yet, how its working...

You try to use rgmixedCG for all shifts in a given monomial?

kostrzewa:
For now, yes. This also makes sense for production purposes, since we often have the two lightest shifts on the coarsest timescale, while the other partial fractions are grouped by threes or fours, where the benefit of using shift by shift solves is not so obvious.

urbach:
okay, I see.

kostrzewa:
@sbacchio @Finkenrath @urbach 
So this is my current solution for solving the multishift systems shift by shift. Choosing a different solver (see solve_mms_nd function) should be rather trivial at this point. It works, but it is not faster, not even for two shifts. The mixed ND solver is about a factor of 1.6 faster than CGMMS, which thus does not compensate for the number of systems to be solved. Of course, once DDalphaAMG is available, this should be beneficial for the smallest shifts.

kostrzewa:
One could now think about taking the lightest or two lightest shifts out of the system and solve these using rgmixedcg, keeping a (more expensive) approximation with more shifts on one of the lower time scales. For this, one would have to study the individual forces though, I believe. One would probably also have to try to estimate which kind of splitting would be optimal. I don't think we can reduce the order, which is already lower than CLS seems to use.

kostrzewa:
I have been running a test of this on a 16c32 lattice to see if it causes any biases. Given the small lattice volume, it is a bit difficult to get the kind of statistical precision that would be required for a conclusive test, but the indication seems to be that the two runs are compatible:

Run with shift-by-shift in all ND*RAT monomials:
```
   L  T    kappa  csw   mul muh musigma mudelta N.online N.plaq skip
1 16 32 0.140065 1.74 0.003   0  0.1408  0.1521     3033   3077  250
                    val         dval     tauint    dtauint Wopt
mpcac_fit  2.009475e-04           NA         NA         NA   NA
mpcac_mc  -1.263659e-05 3.354550e-04  4.7928750 0.93257708   33
mpi        1.520504e-01 2.281873e-03  0.5738911 0.03565074    3
fpi        4.750034e-02 6.998057e-04  0.8957561 0.07700910    6
P          5.408100e-01 5.115257e-05 11.4906936 3.10007378   67
dH         3.154457e-03 1.779947e-03  0.4643404 0.01703494    1
expdH      1.001947e+00 1.758800e-03  0.4629396 0.01699503    1
mineval              NA           NA         NA         NA   NA
maxeval              NA           NA         NA         NA   NA
CG.iter              NA           NA         NA         NA   NA
accrate    9.603639e-01 3.462737e-03  0.4846341 0.01760439    1
```

Run with CGMSS in all ND*RAT monomials:
```
   L  T    kappa  csw   mul muh musigma mudelta N.online N.plaq skip
1 16 32 0.140065 1.74 0.003   0  0.1408  0.1521     1529   1573  250
                   val         dval    tauint    dtauint Wopt
mpcac_fit 0.0004999093           NA        NA         NA   NA
mpcac_mc  0.0004830610 2.687640e-04 1.7922657 0.31366429   13
mpi       0.1523535264 4.045645e-03 0.6864417 0.06856383    4
fpi       0.0481764620 1.277655e-03 1.1455753 0.14770256    7
P         0.5407039315 4.298916e-05 4.7628867 1.19418715   29
dH        0.0065529396 2.439074e-03 0.4645563 0.02383029    1
expdH     0.9982449711 2.305189e-03 0.4556316 0.02347299    1
mineval             NA           NA        NA         NA   NA
maxeval             NA           NA        NA         NA   NA
CG.iter             NA           NA        NA         NA   NA
accrate   0.9644218551 5.178730e-03 0.6146579 0.06107673    4
```

Keep in mind that the former is almost a factor of two longer than the latter. This is due to a scheduling policy on the machine, which effectively prevented the second run from running for the past three weeks or so.

kostrzewa:
This will be merged via the DDalphaAMG_nd changes, closing this pull-request.

PR #338, opened 2016-08-08
read_spinor: add new flag 'DisableSourceIOChecks' 

kostrzewa:
Allow the user to disable the IO check for the reading of sources, 
independently from the gauge IO checks. 

'DisableGaugeIOChecks' parameter added for consistency and
'DisableIOChecks' retained for backwards compatibility. The latter
disables only gauge IO checks, as it used to be.

### DIFF
--- a/read_input.l
+++ b/read_input.l
@@ -124,8 +124,6 @@ static inline void rmQuotes(char *str){
   int return_check_flag, return_check_interval;
   int gauge_precision_read_flag;
   int gauge_precision_write_flag;
-  int g_disable_IO_checks;
-  int g_disable_src_IO_checks;
   int gmres_m_parameter, gmresdr_nr_ev;
   int reproduce_randomnumber_flag;
   double stout_rho;
--- a/read_input.l
+++ b/read_input.l
@@ -125,6 +125,7 @@ static inline void rmQuotes(char *str){
   int gauge_precision_read_flag;
   int gauge_precision_write_flag;
   int g_disable_IO_checks;
+  int g_disable_src_IO_checks;
   int gmres_m_parameter, gmresdr_nr_ev;
   int reproduce_randomnumber_flag;
   double stout_rho;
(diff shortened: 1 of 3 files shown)


read_input.l
```diff
@@ -125,6 +125,7 @@ static inline void rmQuotes(char *str){
   int gauge_precision_read_flag;
   int gauge_precision_write_flag;
   int g_disable_IO_checks;
+  int g_disable_src_IO_checks;
```
urbach: why is this here and in `global.h`?

kostrzewa: I'm following the same pattern as we follow for many other globals, as far as I know. To be honest, I'm a bit confused because it's pretty inconsistent. Some globals are declared in ```global.h```, some are additionally declared in ```read_input.h``` as "extern" (always, not via macro) and some are in addition declared in ```read_input.l``` (without extern). As far as I can see, only ```read_input.l``` now includes ```global.h``` with the ```INIT_GLOBALS``` PP constant defined. As a result, I think the declarations in ```global.h``` and ```read_input.l``` are now equivalent, such that one of them will be ignored. Similarly, including both ```read_input.h``` and ```global.h``` should have the same effect, unless ```INIT_GLOBALS``` is defined at the point of inclusion.

In conclusion,this should be cleaned up across the board, I think.

kostrzewa: So as far as I can tell, one need only remove the declarations in read_input.l and everything should work as required. We should probably discuss if we want to make an adjustment to the naming scheme.

kostrzewa:
I would like to only merge this after the DDalphaAMG interface has been merged because there will be one conflict.

kostrzewa:
Updated this to modify the "fix" for the source io check introduced in the DDalphaAMG branch.

Note @sbaccio, @Finkenrath : this pull-request will add an additional input parameter ```DisableSourceIOChecks``` to disable the IO check for input / initial guess spinors.

PR #339, opened 2016-09-07
read_message(), forgotten instance of MPI -> TM_USE_MPI

kostrzewa:


kostrzewa:
It might be that the LEMON problem that we observed on Jureca may have originated from this, although I'm not sure.

PR #340, opened 2016-09-29
monitor_forces: output 8 digits for force, otherwise large but steady…

kostrzewa:
… forces don't fluctuate, making analysis awkward


kostrzewa:
@urbach: would appreciate a merge of this at some point, cheers

PR #343, opened 2016-10-20
Merging DDalphaAMG interface to etmc/tmLQCD master

sbacchio:


sbacchio:
I cannot solve the conflict, I don't have write access.
@urbach @kostrzewa you should complete the merge.

update_tm.c
```diff
@@ -80,7 +83,7 @@ int update_tm(double *plaquette_energy, double *rectangle_energy,
   /* Energy corresponding to the Gauge part */
   double new_plaquette_energy=0., new_rectangle_energy = 0.;
 
-  /* Energy corresponding to the Momenta part */
+  /* Energy correspondingupdate_tm to the Momenta part */
```
kostrzewa: spurious replacement

kostrzewa:
The conflict should be resolved on the requesting side.

sbacchio:
you mean I should before merge the etmc master to Finkenrath master and then do the pull request?

kostrzewa:
> you mean I should before merge the etmc master to Finkenrath master and then do the pull request?

Yes, that's how one would usually do it.

urbach:
> you mean I should before merge the etmc master to Finkenrath master and then do the pull request? 

yes...

> 
> -- 
> https://github.com/etmc/tmLQCD/pull/343#issuecomment-255077348
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
You can keep the pull request open, any changes that you commit to Finkenrath/tmLQCD/master will automatically propagate here.

solver/solver_params.h
```diff
@@ -56,6 +56,8 @@ typedef struct {
      where the maximum is over the iterated residuals since the last update */  
   float mcg_delta; 
 
+  /*  */
```
kostrzewa: not sure what happened here

sbacchio: I don't remember to do that

operator/D_psi.h
```diff
@@ -35,12 +35,11 @@ void Block_H_psi(block * blk, spinor * const rr, spinor * const s, const int eo)
 void Block_D_psi_32(block * blk, spinor32 * const rr, spinor32 * const s);
 void Block_H_psi_32(block * blk, spinor32 * const rr, spinor32 * const s, const int eo);
 
-
-
-/* csw=0 version*/
-void Dtm_psi(spinor * const P, spinor * const Q);
-/* csw>0 version*/
-void Dsw_psi(spinor * const P, spinor * const Q);
+// TODO: this funtions are just defined for SSE!
```
kostrzewa: @urbach Is this a problem with the way this is set up now?

io/spinor_read.c
```diff
@@ -129,10 +129,12 @@ int read_spinor(spinor * const s, spinor * const r, char * filename, const int p
     }
   }
 
-  if (!DML_read_flag) {
-    fprintf(stderr, "LIME record with name: \"scidac-checksum\", in gauge file %s either missing or malformed.\n", filename);
-    fprintf(stderr, "Unable to verify integrity of gauge field data.\n");
-    return(-1);
+  if (!g_disable_IO_checks){
```
kostrzewa: I have a solution for this waiting to be merged here https://github.com/etmc/tmLQCD/pull/338

I just need to fix something related to the inconsistent definitions of global variables.

kostrzewa: In my solution, sources / propagators and gauges have separate IO_check flags, such that disabling one does not mean that the other is disabled and vice versa.

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,563 @@
...
+#include "read_input.h"
+#include "DDalphaAMG.h"
+#include "linalg_eo.h"
+#include "operator/D_psi.h"
+#include "operator/tm_operators.h"
+#include "operator/clovertm_operators.h"
+
+//Enable to test the solution. It cost an application more of the operator. 
+#define MGTEST
+
```
kostrzewa: Just to let you know, I would like to move some of these ```mg_``` parameters into the globals, such that they can be reused between different MG implementations. Is that okay?

io/spinor_read.c
kostrzewa: For now I would say, let's pull this in. Just wanted to let you know that there will be an additional input parameter for this behaviour.

DDalphaAMG_interface.c
kostrzewa: Also, would we want ```MGTEST``` to be compile-time configurable?

solver/solver_params.h
kostrzewa: Could you still remove that line please?

operator/D_psi.h
sbacchio: Sorry, I did some changes which maybe are stupid.
The point was that when I've interfaced all the operator functions with DDalphaAMG (look at the MG_solve function in the interface), I tried to interface these ones as well and I had errors compiling because they are defined just in the SSE part and not in a general way. So i comment out them in order to be sure that these functions won't be called in the MG before being interfaced.

DDalphaAMG_interface.c
sbacchio: Fine

kostrzewa: Regarding ```MGTEST```, it's entirely your call I would say.

operator/D_psi.h
kostrzewa: I remember our discussion about it. I think these few lines should be surrounded by 

```
#if (defined SSE || defined SSE2 || defined SSE3 )
  void ....
#endif
```

kostrzewa: What do you think, @urbach?

solver/solver_params.h
sbacchio: done, I check if there is something more before commiting

io/spinor_read.c
sbacchio: ok

DDalphaAMG_interface.c
sbacchio: Do you have a way to compile a "debug" version where this flag is enabled?
If not, I think at the moment is still useful, because we didn't tested all the cases yet. So it's good to know when it fails.

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,563 @@
...
+int mg_initialized=0;
+int mg_setup_iter=5;
+int mg_coarse_setup_iter=3;
+int mg_update_setup_iter=1;
+int mg_omp_num_threads=0;
+int mg_Nvec=24;
+int mg_lvl=3;
+int mg_blk[4] = {0, 0, 0, 0};
+int mg_mixed_prec=1;
+double mg_setup_mu = 0./0.; //setting to NaN
```
kostrzewa: So according to http://stackoverflow.com/a/570694 this may indeed be unsafe as soon as the compiler does not respect the IEEE standard for floats/doubles. Would a scaling factor perhaps be safer? That way, it can be ```1.0``` by default, there is one fewer if condition below and the user will not be confused by the fact that this is actually ```2*kappa*mu``` rather than just the twisted mass itself.

DDalphaAMG_interface.c
kostrzewa: At least in principle, the build harness has a debug version, but I think it's quite broken at this point. Let's leave it as it is right now and earmark it for a possible adjustment in the future, would you agree?

DDalphaAMG_interface.c
sbacchio: ok, what about naming the parameter MGSetup2KappaMu (so we don't get confused) and then use the scale factor to save the value - i.e scale_factor = MGSetup2KappaMu / 2KappaMu?

DDalphaAMG_interface.c
sbacchio: Ok, I will add a TODO there for the future.

DDalphaAMG_interface.c
kostrzewa: I understand that it makes a particular type of parameter scan trickier: when you want to fix the setup mu and vary the target mass.

sbacchio: Yes indeed, asking in the input file for a scale factor is annoying because you have to adapt it everytime. Better to give directly a `MGSetup2KappaMu`. 
The only case not covered is when you don't have a mu (pure wilson), but you may want to do a setup with mu. It's something which we have tested and do not give advantages, so there is no reason for doing it...
Concluding, do you approve `MGSetup2KappaMu` stored in a scale factor mg_setup_mu_factor?

kostrzewa: Sure, if we can get away without ``` f != f``` as a test for NaN (to check if the user has even set this parameter), then this will be safe.

sbacchio: done

operator/D_psi.h
sbacchio: At the moment I've added the `#if (defined SSE || defined SSE2 || defined SSE3 )` and the check in the interface.
I wait for a confirmation on this point and then I commit all the new changes.

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,563 @@
...
+  return mg_status.success;
+}
+
+void MG_init()
+{
+  mg_init.comm_cart=g_cart_grid;
+  mg_init.Cart_rank=Cart_rank;
+  mg_init.Cart_coords=Cart_coords;
+  
+  mg_init.global_lattice[0]=T*N_PROC_T;
```
kostrzewa: @urbach: would you recommend using ```g_nproc_[x,y,z,t]``` rather than the input values?

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,563 @@
...
+  mg_params.conf_index_fct=conf_index_fct;
+  mg_params.vector_index_fct=vector_index_fct;
+  
+  /* in DDalphaAMG
+   * Printing level:
+   *  -1: silent (errors or warnings)
+   *   0: minimal //default
+   *   1: verbose
+   */
+  if(verbose) {
```
kostrzewa: should this depend on ```g_debug_level``` rather than verbose? It seems that the verbosity status gets lost after the first trajectory in the HMC...

sbacchio:
I've pushed a new version. I've added the changes we have agreed and as well some improvements in the heatbath and acceptance step. There, indeed, the inversion of the squared operator was called in the MG, when the non squared is enough.

kostrzewa:
@urbach , could you look into this?
> @kostrzewa @urbach: please, check if the use of the chrono function is still fine.. I've tried to understand how they work but I'm not sure if I applied the change correctly there.

I'm also still not sure if my suggestion regarding the SSE versions of D_psi (twisted mass / [tm]clover) were sensible...

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,576 @@
...
+  
+  if(!success) {
+    if( g_proc_id == 0 )
+      printf("ERROR: solver didn't converge after two trials!! Aborting... \n");
+    //TODO: handle abort
+    DDalphaAMG_finalize();
+    MPI_Barrier(MPI_COMM_WORLD);
+    MPI_Finalize();
+    exit(0);
+  }
```
kostrzewa: Since you are terminating tmLQCD here after an error, the exit status should be non-zero.

operator/D_psi.h
```diff
@@ -35,12 +35,12 @@ void Block_H_psi(block * blk, spinor * const rr, spinor * const s, const int eo)
 void Block_D_psi_32(block * blk, spinor32 * const rr, spinor32 * const s);
 void Block_H_psi_32(block * blk, spinor32 * const rr, spinor32 * const s, const int eo);
 
-
-
+#if (defined SSE || defined SSE2 || defined SSE3 )
```
sbacchio: ok, it will be in the next commit.

kostrzewa:
I noticed a problem which I don't quite understand. When I do multiple inversions (random samples or some point sources, say), I need to set ```MGUpdateSetupIter=0```, otherwise ```MG_pre_solve``` will call ```MG_update_mu()``` and ```DDalphaAMG_update_setup()``` between each inversion even though the gauge configuration was not changed...

I believe this could be fixed by setting the default for ```mg_update_setup_iter``` to 0, rather than the current 1. If that is undesirable (because it will complicate the HMC logic), then the interface should know that the gauge configuration has not been changed...

kostrzewa:
Maybe the big if condition in ```MG_pre_solve``` contains a logic bug, not sure.

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,576 @@
...
+  double dtau= mg_tau-gauge_tau;
+  // Checking if:
+  //  mg_update_setup < mg_update_setup_iter : maybe you want to do more iteration at this run
+  //  mg_dtau_update < dtau  : regular condition for update of setup
+  //  mg_dtau_update < -dtau : during reversability check dtau is negative!
+  //  mg_dtau_update == 0.0  : updating at every change of configuration -> valid as well if configuration changed outside the HMC
+  //  mg_rho_update < 0.0    : parameter ignore
+  //  mg_rho_update == rho   : updating only if this condition and the others are satisfied
+  if ( mg_do_setup == 0 && mg_update_setup < mg_update_setup_iter && ( mg_dtau_update < dtau+1e-6 || mg_dtau_update < -dtau-1e-6 || (mg_dtau_update==0.0 && mg_update_gauge==1)) &&
+       (mg_rho_update < 0.0 || mg_rho_update == g_mu3))
```
sbacchio: yes the problem is here.. `(mg_rho_update < 0.0 || mg_rho_update == g_mu3)` should be `(mg_rho_update >= 0.0 && mg_rho_update == g_mu3)`

kostrzewa: I see, that makes sense. Might fixing this break the reweighting monomial?

kostrzewa: > I don't understand what you mean.

The reweighting monomial based on the DDalphaAMG branch worked on by @gbergner https://github.com/gbergner/tmLQCD/tree/ReweightingDDalphaAMG

kostrzewa:
For ```MGMixedPrecision = yes```, I experience rather frequent "converged to the wrong solution" errors. I also don't really see a performance benefit of using a mixed-precision solver, or was your experience different? (This is using the latest version of DDalphaAMG which includes the global reduction)

DDalphaAMG_interface.c
sbacchio: yes, but Jacob was telling me that there was a problem when the even odd implementation was used.. Now that problem should be solved.

sbacchio:
> For `MGMixedPrecision = yes`, I experience rather frequent "converged to the wrong solution" errors. I also don't really see a performance benefit of using a mixed-precision solver, or was your experience different? (This is using the latest version of DDalphaAMG which includes the global reduction)

Sincerely I don't know anymore.. At the beginning I was noticing an improvement but I understand that now maybe it's not anymore. And moreover with the 1+1 implementation the mixed precision solver is not working well at all (but I think due still to some bug). Anyway after this work I will review the mixed precision performance.

sbacchio:
Can we please go ahead with the merge? I've some new changes to push but I was waiting first the review of @urbach specially about the changing in `monomial/*` where the chronological solver is involved.

kostrzewa:
@urbach should we just proceed?

monomial/cloverdet_monomial.c
```diff
@@ -223,13 +223,20 @@ double cloverdet_acc(const int id, hamiltonian_field_t * const hf) {
...
-  chrono_guess(mnl->w_fields[1], mnl->pf, mnl->csg_field, mnl->csg_index_array,
-               mnl->csg_N, mnl->csg_n, N, mnl->Qsq);
   g_sloppy_precision_flag = 0;
-  mnl->iter0 += solve_degenerate(mnl->w_fields[0], mnl->pf, mnl->solver_params, mnl->maxiter, mnl->accprec,  
-                                 g_relative_precision_flag, VOLUME/2, mnl->Qsq, mnl->solver); 
-  mnl->Qm(mnl->w_fields[0], mnl->w_fields[0]);
 
+  if( mnl->solver == MG ){
+      chrono_guess(mnl->w_fields[1], mnl->pf, mnl->csg_field, mnl->csg_index_array,
+		   mnl->csg_N, mnl->csg_n, N, mnl->Qp);
```
kostrzewa: @urbach could you please check whether this is still correct now? It would be great to pull this in today, such that @sbacchio can continue with the 1+1 pull-request. The 1+1 code will need testing, for which I'd like to do another run in Zeuthen.

monomial/cloverdet_monomial.c
```diff
@@ -223,13 +223,20 @@ double cloverdet_acc(const int id, hamiltonian_field_t * const hf) {
...
   }
 
-  chrono_guess(mnl->w_fields[1], mnl->pf, mnl->csg_field, mnl->csg_index_array,
-               mnl->csg_N, mnl->csg_n, N, mnl->Qsq);
   g_sloppy_precision_flag = 0;
-  mnl->iter0 += solve_degenerate(mnl->w_fields[0], mnl->pf, mnl->solver_params, mnl->maxiter, mnl->accprec,  
-                                 g_relative_precision_flag, VOLUME/2, mnl->Qsq, mnl->solver); 
-  mnl->Qm(mnl->w_fields[0], mnl->w_fields[0]);
 
+  if( mnl->solver == MG ){
```
kostrzewa: @urbach @sbacchio 
It would be nicer to not have this conditional here, that's what ```solve_degenerate``` was introduced for. But the chronological solver clearly complicates matters. My guess is that the matrix passed to ```chrono_guess``` should still be the squared one.

DDalphaAMG_interface.c
```diff
@@ -0,0 +1,576 @@
...
+	    f == Qsw_minus_psi || // Gamma5 - Schur complement with minus mu
+	    f == Qtm_pm_psi ||    //          Schur complement squared
+	    f == Qsw_pm_psi ) {   //          Schur complement squared
+    if( N != VOLUME/2 && g_proc_id == 0 )
+      printf("WARNING: expected N == VOLUME/2 for the required operator in MG_solve. Continuing with N == VOLUME\n");
+  }
+  else if ( f == D_psi ||         //          Full operator    with plus mu
+	    f == Q_plus_psi ||    // Gamma5 - Full operator    with plus mu 
+	    f == Q_minus_psi ||   // Gamma5 - Full operator    with minus mu
+#if (defined SSE || defined SSE2 || defined SSE3 )
```
kostrzewa: These are probably not necessary, as discussed. When SSE is enabled and the clover parameter is non-negative, the correct operator will be called by ```D_psi```.

sbacchio: yes, I've already fixed it, but it's in the next commit. I wanted to review the cronosolver first so I commit everything together.

monomial/cloverdet_monomial.c
sbacchio: no the difference is exactly there: in the heatbath and acceptance step we don't need the inversion of the squared operator and indeed what you do is to invert the squared operator and apply directly the D(-mu) to retrieve the unsquared:
`-  mnl->iter0 += solve_degenerate(mnl->w_fields[0], mnl->pf, mnl->solver_params, mnl->maxiter, mnl->accprec,  
￼ -                                 g_relative_precision_flag, VOLUME/2, mnl->Qsq, mnl->solver); 
￼ -  mnl->Qm(mnl->w_fields[0], mnl->w_fields[0]);`
but with the MG we can invert the unsquared directly and we don't need to do two inversions in this case.
so what I do here is to give at the solver the unsquared one: `mnl->Qp`.

Now I'm not sure about the cronological solver, because in both cases when I call `chrono_guess` and `chrono_add_solution` I give the inversion of the unsquared operator. So I think it's right how I did it.

For clarification, check what is done in `det_monomial.c`. There as well the non squared operator is inverted. And I think the cronological solver is used in the same way.

kostrzewa: Yes, you are right, in the heatbath/accept steps, the square is taken into account in the scalar product. As for ```chrono_guess```, the problem is that now in the derivative, the squared operator is passed, while here the unsquared operator is given. Also, ```chrono_guess``` assumes that the operator is hermitian...

sbacchio: ah I didn't know about the hermiticity required to the operator. and is it actually in that way? because in `det_monomial.c` you give the unsquared operator to the chronological solver.

Anyway I think it's useful for saving time to don't invert twice in the heatbath/accept steps, but I don't know what changes has to be done for the chronological solver since I don't know about it. So my conclusion would be to leave it in this way and give to who knows the task of fixing the chronological solver.

urbach: Currently indeed the condition that the operator must be hermitian is hard-coded in the chronological solver part. This condition is violated more or less strongly by the twisted mass term. tmLQCD is slightly inconsistent there because the BiCGstab was originally only be used without the twisted mass term, simply because it does not converge reliably otherwise. I suppose it has actually never been used.

In principle its a rather simple fix to drop the condition of hermiticity, as I think this condition is not needed for LU solve, right? But I have to check whether or not the functional makes sense in that case. And, one would need multiplications with gamma5, I think.

So, I have a slight preference to merge as it is now and to keep an issue open for the chronological solver.

sbacchio: Yes the hermiticity condition is not required for the LU solver. I had to do similar changes in DDalpha AMG as well, where the Cholesky decomposition was used and it require the operator to be hermitian. 

Ok, I agree with keeping this issue open.

Can you just check if I changed well in the acceptance and heat-bath steps what necessary in order to compute just one inversion with the multigrid? i.e. operator given to solve_degenerate and handling of the solution after.

urbach: Yes, I think its okay. In particular, it will work when the chrono solver is switched off, as far as I can tell.

kostrzewa:
So I think we can pull this in now, unless there are any more bugfixes that you would like to get in @sbacchio @Finkenrath ?

sbacchio:
Yes, I've still some TODO on the list, but no bug fix at the moment. You can merge for me.

kostrzewa:
Okay, I'm merging now.

Issue #346, opened 2016-11-28
DDalphaAMG breaks read_input.l


kostrzewa:
@sbacchio @Finkenrath @urbach 
We overlooked a little something when we merged the DDalphaAMG branch. The lines in ```read_input.l``` which set MG parameters should be enclosed in ```#ifdef```, such that the code can be linked when the ```DDalphaAMG``` interface is not compiled. Alternatively, the MG parameters can be provided globally, also when ```DDalphaAMG``` is disabled.

urbach:
can you provide a pull request for the required change?

> @sbacchio @Finkenrath @urbach 
> We overlooked a little something when we merged the DDalphaAMG branch. The lines in ```read_input.l``` which set MG parameters should be enclosed in ```#ifdef```, such that the code can be linked when the ```DDalphaAMG``` interface is not compiled. Alternatively, the MG parameters can be provided globally, also when ```DDalphaAMG``` is disabled.
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/etmc/tmLQCD/issues/346
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach
URL: http://www.carsten-urbach.eu

sbacchio:
Ah I've been too sloppy.
Sorry, I did a mess. I will fix it.

kostrzewa:
This has been resolved by #347, closing.

PR #347, opened 2016-11-28
Fixing compilation without DDalphaAMG

sbacchio:
In response to issue #346.

@kostrzewa @urbach What do you think about this solution?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,5 +18,4 @@ invert
 offline_measurement
 lib/
 benchmark
-config.status
-config.h
+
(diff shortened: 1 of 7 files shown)


kostrzewa:
@urbach could you take a look at this please? I'm okay with the solution proposed by @sbaccio.

kostrzewa:
I might, at some point in the future, provide a pull-request which moves the mg parameters out of DDalphaAMG and into a global context, thereby providing a uniform parameter set for different MG implementations, as far as possible. (The set of paramters might be extended in the future).

urbach:
yes, this is fine with me, should I merge it in?

kostrzewa:
Sure, please go ahead. Thanks!

PR #350, opened 2016-12-29
Mybranch

marcuspetschlies:
missing includes and files for support of ARPACKCG and EXACTDEFLATEDCG solver; still need to add entries in invert_eo etc.


kostrzewa:
I guess this will never be merged, closing.

PR #353, opened 2017-01-30
Bug fix in correction term for rational approximation.

sbacchio:
The series expansion of `B =  (1+Z)^{1/4}` had the wrong factor for the order `Z^3` in both, code and documentation.
The series expansion of `(1+Z)^{-1/2}` was wrong in the documentation but correctly implemented in the code.


sbacchio:
This bug should not be relevant for the ensemble generated; because from what I see in the 16c32 ensemble, the Z^3 term was not relevant in the series expansion.
It starts to be relevant for the 64c128 ensemble.

kostrzewa:
And even so, the mismatch of the coefficient, 6/128, is probably so small, that its effect on the correction monomial will be almost certainly irrelevant. @urbach: I guess this can be merged in right away, yes? How do we proceed with current simulations, continue with the 1/122 factor and earmark the computation of reweighting factors (just to make sure that the effect is really negligible) or switch, mid-simulation, to 1/128 ?

urbach:
> And even so, the mismatch of the coefficient, 6/128, is probably so
  small, that its effect on the correction monomial will be almost
  certainly irrelevant. @urbach: I guess this can be merged in right
  away, yes? How do we proceed with current simulations, continue with
  the 1/122 factor and earmark the computation of reweighting factors
  (just to make sure that the effect is really negligible) or switch,
  mid-simulation, to 1/128 ? 

yes, we can merge. I'd change on the large volumes, since they are not
yet really equilibrated.

> 
> -- 
> https://github.com/etmc/tmLQCD/pull/353#issuecomment-276949439
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
> yes, we can merge. I'd change on the large volumes, since they are not yet really equilibrated.

Sure, on the runs which are not equilibrated should switch to the correct factor asap. But what about the other runs?

PR #355, opened 2017-02-07
Improvements to correction monomial

sbacchio:
Saved some applications of Z in both heatbath/acc step.
Improved the check of next order correction in both heatbath/acc step (which we should discuss if it was a bug; read following).
Added documentation for guiding the changes.

Everything tested and gives same results of before.
A test is running on supermuc for the ensemble 16c Nf=1+1 with changes included, using just DDalphaAMG. You can find it in /gpfs/work/pr74yo/di49saj/HMC_MG_nd_only_test.

@urbach @kostrzewa @Finkenrath 
-------------------

The improvements in the checks for next order correction are the following; 
we should discuss if it was a bug.

The old check for the order `n`th was
  `| Z^n R |^2 < acceptance^2`
which reads as `R^\dagger Z^{2n} R `.
The new check is
  `[ c_n R^\dagger Z^n R ]^2 < acceptance^2`
which is actually the correction to the energy given by the order `n`th.

About the old check we can say
  `| Z^n R |^2 < |Z|^{2n} |R|^2`
and about the new instead
  `[ R^\dagger Z^n R ]^2 < |Z|^{2n} |R|^4`.
So the upper limits differ of a factor |R|^2 between old and new.
The difference between the two checks is more relevant as much the lattice size increases, since |R|~V.

Indeed what we noticed now on the lattice 64c, is that the new check include terms of higher order compared to the old check (e.g. with the new check the terms of order Z^3 are included in the series; while with the old check they weren't).
On smaller lattices, as the 16c, there is no difference between old and new check because both of them cut the series at order Z^2.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -50,7 +50,7 @@ MODULES = read_input gamma measure_gauge_action start \
 	geometry_eo invert_overlap aligned_malloc \
 	prepare_source chebyshev_polynomial_nd Ptilde_nd  \
 	reweighting_factor_nd rnd_gauge_trafo \
-        update_momenta integrator  phmc \
+        update_momenta update_momenta_fg integrator  phmc \
 	little_D block operator \
 	temporalgauge spinor_fft X_psi P_M_eta \
 	jacobi fatal_error invert_clover_eo gettime @SPI_FILES@ \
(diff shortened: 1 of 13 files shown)


kostrzewa:
I need to sit down and work through this on paper, I hope I find time for it soon...

kostrzewa:
Note that the conflicts should be resolved and threading enabled before the new integrator can be merged.

PR #356, opened 2017-02-09
Some bug fix about MG_update.

sbacchio:
In response to the issues reported by @kostrzewa:

> - when using the DDlpahaAMG interface in invert, the MG setup is updated when the source changes or the flavour is switched, unless I set MGUpdateSetupIter = 0.
> 
> - Worse still, when processing multiple gauge configurations in one job, I've noticed that the DDalphaAMG interface will not update its copy of  the gauge configuration between gauge configurations. This is true whatever value I set MGUpdateSetupIter to. After the new gauge configuration has been read in, it will
> a) do MGUpdateSetupIter setup iterations (without updating the gauge field first), if MGUpdateSetupIter > 0
> b) attempt to do one inversion, notice that the result is wrong, and restart.
>
> Only after this restart will the gauge field be updated and the setup1 redone fully.

1. Fixed reset of multigrid when several measurements are done with invert or offline_measurments.
2. Changed position MG_reset when configuration is rejected. 
3. Fixed update of setup.

@kostrzewa can you please check if now everything works?

### DIFF
--- a/DDalphaAMG_interface.c
+++ b/DDalphaAMG_interface.c
@@ -210,8 +210,10 @@ static int MG_pre_solve( su3 **gf )
   //  mg_dtau_update == 0.0  : updating at every change of configuration -> valid as well if configuration changed outside the HMC
   //  mg_rho_update < 0.0    : parameter ignore
   //  mg_rho_update == rho   : updating only if this condition and the others are satisfied
-  if ( mg_do_setup == 0 && mg_update_setup < mg_update_setup_iter && ( mg_dtau_update < dtau+1e-6 || (mg_dtau_update==0.0 && mg_update_gauge==1) ||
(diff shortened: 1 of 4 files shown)


kostrzewa:
Yes, this seems to work as expected. Haven't tested in the HMC, however. I guess that should also still operate correctly. Should I pull it in?

sbacchio:
I checked it again and everything looks fine to me.

We don't have at the moment a HMC to try it, because we don't want to change the executables on the way. Anyway I guess it should work correctly.

Can you please have a look to my other pull request when you have time?

kostrzewa:
Of course, it's high on my list of priorities, but it's a little bit more work to look into and I'm currently rather busy...

sbacchio:
thanks! :)

PR #358, opened 2017-03-16
further refactor bechmark - DO NOT MERGE YET

kostrzewa:
Some changes to benchmark to clean up the main function in order to add benchmarks of the single precision operators. 

Further work will be done to do the same also for Dpsi.


kostrzewa:
Please do not merge yet.

PR #367, opened 2017-04-22
Add Travis CI support

martin-ueding:
To actually build on Travis CI, go to https://travis-ci.org/etmc/tmLQCD
and log in with an account having admin rights on that repository. There
should be a switch to enable building. Then it is automatically
integrated with GitHub.

Currently it uses one of the HMC input files to run it. It would be even
better to use some proper unit test. The current usage is a “smoke test”
(as in “Does it smoke when it is turned on?”) only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,24 @@
+########################################################################
+#
+# Copyright (C) 2017 Martin Ueding <dev@martin-ueding.de>
+#
+# This file is part of tmLQCD.
+#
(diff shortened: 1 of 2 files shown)


urbach:
Okay, I merge this in. We have to see how useful it actually is in this form.

urbach:
I think I have also activated it now.

martin-ueding:
One has to add more useful tests in `travis-ci.sh`. Perhaps one can also build it with more fancy options to test-drive a larger part of the codebase.

PR #370, opened 2017-05-02
quda_interface: support current inv_param setup for mu and clover term

kostrzewa:
support twisted mass flavour choice in current quda develop branch


kostrzewa:
@urbach 
it would be great if this could be pulled in soon as it is blocking JSC for their benchmarks as I somehow forgot to push these changes earlier with all the stress.

PR #378, opened 2017-05-10
qphix_interface: fix checkerboard convention

kostrzewa:
now both test_Dslash and the inversion work correctly for Wilson


kostrzewa:
Let's see if the build goes through now.

kostrzewa:
@martin-ueding thanks a lot for setting up travis, even in this simple setup, it is very useful

qphix_interface.cpp
```diff
@@ -766,20 +766,20 @@ int invert_eo_qphix_helper(spinor *const tmlqcd_even_out, spinor *const tmlqcd_o
...
                          1);               // non-conjugate
   WilsonDslash->dslash(qphix_in_prepared,  // out spinor
                        qphix_buffer,       // in spinor
-                       u_packed[0],        // gauge field on target cb
+                       u_packed[1],        // gauge field on target cb
                        1,                  // non-conjugate
-                       0);                 // target cb == odd
-  QPhiX::aypx(0.5, qphix_in[0], qphix_in_prepared, geom, n_blas_simt);
+                       1);                 // target cb == odd
+  QPhiX::aypx(0.5, qphix_in[1], qphix_in_prepared, geom, n_blas_simt);
```
martin-ueding: This looks good, one has be extra careful because the function is `aypx(a, x, y)`. But that looks good 👍

qphix_interface.cpp
```diff
@@ -805,15 +805,15 @@ int invert_eo_qphix_helper(spinor *const tmlqcd_even_out, spinor *const tmlqcd_o
     //   M M^dagger qphix_buffer = qphix_in_prepared
     // here, that is, isign = -1 for the QPhiX CG solver.
     // After that multiply with M^dagger:
-    //   qphix_out[0] = M^dagger M^dagger^-1 M^-1 qphix_in_prepared
+    //   qphix_out[1] = M^dagger M^dagger^-1 M^-1 qphix_in_prepared
     (*SolverQPhiX)(qphix_buffer, qphix_in_prepared, RsdTarget, niters, rsd_final, site_flops,
                    mv_apps, -1, verbose);
-    (*FermionMatrixQPhiX)(qphix_out[0], qphix_buffer, /* conjugate */ -1);
+    (*FermionMatrixQPhiX)(qphix_out[1], qphix_buffer, /* conjugate */ -1);
```
martin-ueding: I do not understand why the `isign = -1` for both. That looks like we have `M^\dagger M^\dagger x = b`. If it is `M M^dagger qphix_buffer = qphix_in_prepared`, then wouldn't the solver solve with `isign = 1`?

qphix_interface.cpp
```diff
@@ -766,20 +766,20 @@ int invert_eo_qphix_helper(spinor *const tmlqcd_even_out, spinor *const tmlqcd_o
   // c) Apply AYPX to rescale last result (=y) and add b_o (=x)
 
   DslashQPhiX->A_inv_chi(qphix_buffer,     // out spinor
-                         qphix_in[1],      // in spinor
+                         qphix_in[0],      // in spinor
```
kostrzewa: So, ```cb_even``` and ```cb_odd``` are currently in an anonymous namespace inside the ```tmlqcd``` namespace.  Do you think we should have ```using namespace tmlqcd``` in ```qphix_interface.cpp```?

qphix_interface.cpp
```diff
@@ -60,6 +60,8 @@ extern "C" {
```
kostrzewa: @martin-ueding Could I ask you to go through this please and check that I didn't mess up the magic number of cb_[even,odd] conversion?

qphix_interface.cpp
```diff
@@ -688,7 +688,7 @@ int invert_eo_qphix_helper(spinor *const tmlqcd_even_out, spinor *const tmlqcd_o
     abort();
   } else if (g_mu != 0.0) {  // TWISTED-MASS
     QPhiX::masterPrintf("# Creating QPhiX Twisted Mass Wilson Dslash...\n");
-    const double TwistedMass = -g_mu / (2.0 * g_kappa);
```
kostrzewa: oops, this should probably not have been in here yet

martin-ueding:
I have checked the commit where you have changed the magic numbers to `cb_even` and `cb_odd`. That looks good to me!

PR #382, opened 2017-05-16
make qphix_test_Dslash meaningful

kostrzewa:
* provide support for reading operator type from input file
* implement single application of full operator in qphix interface

### DIFF
--- a/qphix_interface.cpp
+++ b/qphix_interface.cpp
@@ -38,11 +38,14 @@ extern "C" {
 #include "geometry_eo.h"
 #include "gettime.h"
 #include "linalg/convert_eo_to_lexic.h"
+#include "linalg/square_norm.h"
 #include "operator/clovertm_operators.h"
 #include "solver/solver.h"
 #include "solver/solver_field.h"
 #include "solver/solver_params.h"
 #include "update_backward_gauge.h"
+#include "operator_types.h"
+#include "start.h"
 }
 #ifdef TM_USE_OMP
 #include <omp.h>
@@ -57,7 +60,7 @@ extern "C" {
 #include <cfloat>
 #include <cstdlib>
 #include <cstring>
-
+#include <cfloat>
 #include <vector>
 
 using namespace tmlqcd;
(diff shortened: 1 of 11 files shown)


kostrzewa:
Okay, so I think we're ready to finish working on the clover stuff now!

qphix_interface.cpp
```diff
@@ -819,8 +814,10 @@ int invert_eo_qphix_helper(spinor *const tmlqcd_even_out, spinor *const tmlqcd_o
   double end = omp_get_wtime();
 
   uint64_t num_cb_sites = lattSize[0] / 2 * lattSize[1] * lattSize[2] * lattSize[3];
+  // FIXME: this probably needs to be adjusted depending on the operator used
   uint64_t total_flops = (site_flops + (72 + 2 * 1320) * mv_apps) * num_cb_sites;
   QPhiX::masterPrintf("# Solver Time = %g sec\n", (end - start));
+  QPhiX::masterPrintf("# Flops per site = %ld \n", site_flops );
```
kostrzewa: I should probably remove this, I don't understand its value...

martin-ueding: You mean the `total_flops`? In the QPhiX tests, `site_flops` is used to track the number of operations from the BLAS routines and `mv_apps` is the number of fermion matrix operator applications.

martin-ueding:
So the interface for Wilson and Wilson twisted mass is done? That sounds great! 😃 

Do we want to continue with Wilson clover or TM clover?

qphix_interface.cpp
kostrzewa: I meant site flops. Okay, but then the constant multiplying mv_apps is not correct. 1320 is for Wilson only.

kostrzewa:
> So the interface for Wilson and Wilson twisted mass is done? That sounds great! :smiley:

Up to the problem with communication in two space-time directions simultaneously... I was planning to look at the QDP packers to understand this.

kostrzewa:
This is pulled in by #383, closing.

PR #383, opened 2017-05-26
updated qphix_test_Dslash & full 4D parallelisation support

kostrzewa:


qphix_interface.cpp
```diff
@@ -246,91 +276,195 @@ void reorder_gauge_to_QPhiX(QPhiX::Geometry<FT, VECLEN, SOALEN, compress12> &geo
...
+  // while tmlqcd does
+  //   -+t -> -+x -> -+y -> -+z
+  // same as the lattice ordering
+  // The mappingn between the application dimensions is thus:
+  //  tmlqcd_dim(t(0) -> x(1) -> y(2) -> z(3)) = qphix_dim( t(3) -> x(0) -> y(1) -> z(2) )
+  const int change_dim[4] = {1, 2, 3, 0};
 
   // Get the base pointer for the (global) tmlQCD gauge field
-  uint64_t tm_idx = 0;
+  xchange_gauge(g_gauge_field);
```
kostrzewa: For now I would like to keep this here, but perhaps we can think about some meta-information about whether halos need to be updated or not, so as to not waste cycles...

PR #384, opened 2017-06-02
QPhiX support for all degenerate operators in the inverter

kostrzewa:
* full 4D parallelisation
* more meaningfull test structure for operators
* eo spinor packers
* reworked gauge packer
* clover field packers

### DIFF
--- a/qphix_base_classes.hpp
+++ b/qphix_base_classes.hpp
@@ -378,7 +378,7 @@ class Dslash {
     \param[out] out Output spinor \f$ \psi \f$.
     \param[in] in Input spinor \f$ \chi \f$.
     */
-  virtual void A_chi(Spinor *const out, Spinor const *const in, int const isign) = 0;
+  virtual void A_chi(Spinor *const out, Spinor const *const in, int const isign, int const cb) = 0;
 
   /**
     Computes \f$ \psi_\mathrm e = A_\mathrm{ee}^{-1} \chi_\mathrm e \f$.
@@ -386,7 +386,8 @@ class Dslash {
     \param[out] out Output spinor \f$ \psi \f$.
     \param[in] in Input spinor \f$ \chi \f$.
     */
-  virtual void A_inv_chi(Spinor *const out, Spinor const *const in, int const isign) = 0;
+  virtual void A_inv_chi(Spinor *const out, Spinor const *const in, int const isign,
+                         int const cb) = 0;
 
   /**
     Forwarder for the `dslash`.
@@ -418,7 +419,7 @@ class Dslash {
 
     auto tmp = QPhiX::makeFourSpinorHandle(*geom);
     dslash(tmp.get(), psi, u, isign, cb);
-    A_chi(res, tmp.get(), isign);
+    A_chi(res, tmp.get(), isign, cb);
   };
 
   /**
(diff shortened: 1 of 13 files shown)


qphix_base_classes.hpp
```diff
@@ -611,35 +621,51 @@ class WilsonClovTMDslash : public Dslash<FT, veclen, soalen, compress12> {
...
+      clover_product(out, in, inv_clover[cb][1], upstream_dslash.getGeometry());
+    } else {
+      clover_product(out, in, inv_clover[cb][0], upstream_dslash.getGeometry());
+    }
   }
 
   void dslash(Spinor *const res, const Spinor *const psi, const SU3MatrixBlock *const u,
               int const isign, int const cb) override {
-    upstream_dslash.dslash(res, psi, u, inv_clover, isign, cb);
+    upstream_dslash.dslash(res, psi, u, (const FullCloverBlock **)inv_clover[1 - cb], isign, cb);
```
martin-ueding: I am not sure about the checkerboarding. The Dslash will compute `A^{-1}_{ee} D_{eo}` when you have the even as the target checkerboard. Therefore the inverse clover should also be on the even-even. Below for the `achimbdpsi`, we compute `A_{oo} \psi_o - 1/4 D_{oe} [ A^{-1}_{ee} D_{eo} \psi_p ]` where the stuff in the bracket has been computed by the `dslash` function.

If one takes a look into EO linear operator implementation (`wilson.h` in QPhiX), it calls the first Dslash with an opposing sign. So I would think that it should be `inv_clover[cb]` here.

Perhaps use a `static_cast` instead of the C-style cast, but that is not so important.

qphix_base_classes.hpp
```diff
@@ -650,8 +676,8 @@ class WilsonClovTMDslash : public Dslash<FT, veclen, soalen, compress12> {
   double const derived_mu;
   double const derived_mu_inv;
 
-  FullCloverBlock *const clover[2];
-  FullCloverBlock *const inv_clover[2];
+  FullCloverBlock *clover[2][2];
```
martin-ueding: Could we have a comment here describing what the two indices are? Or is that done somewhere else already?

qphix_base_classes.hpp
```diff
@@ -594,10 +604,10 @@ class WilsonClovDslash : public Dslash<FT, veclen, soalen, compress12> {
     signature, the clover term is a member. This means that the user has to construct a new operator
     each time the clover term has changed.
     */
-  CloverBlock *const clover;
+  CloverBlock *clover[2];
```
martin-ueding: Could the comment above reflect that the array index is for checkerboarding?

qphix_interface.cpp
```diff
@@ -458,132 +921,146 @@ void reorder_spinor_from_QPhiX(QPhiX::Geometry<FT, VECLEN, SOALEN, compress12> &
...
 
-  geom.free(packed_gauge_cb0);
-  geom.free(packed_gauge_cb1);
-  geom.free(packed_spinor_in_cb0);
-  geom.free(packed_spinor_in_cb1);
-  geom.free(packed_spinor_out_cb0);
-  geom.free(packed_spinor_out_cb1);
   geom.free(tmp_spinor);
+  for (int cb : {0, 1}) {
+    geom.free(u_packed[cb]);
```
martin-ueding: If you want, you can use the RAII wrappers for the QPhiX data structures which are relatively new in the Geometry class.

qphix_base_classes.hpp
kostrzewa: Good idea

qphix_base_classes.hpp
kostrzewa: You're right, I had this correct at some point and then broke it...

qphix_base_classes.hpp
kostrzewa: Will do

qphix_interface.cpp
kostrzewa: Yes, there's a comment somewhere about that. I didn't switch to these yet because I wanted to think if it would be possible to have the fields persist during calls to the solver and simply update them as required. This, however, is essentially only useful in the HMC and would require a substantial amount of logic to make happen...

kostrzewa:
Okay, I think I've resolved all remaining issues, merging.

Issue #393, opened 2017-07-31
Problems compiling with SSE intrinsics


Irubataru:
I'm having problems compiling tmLQCD with SSE intrinsics on my computer. I run configure with the following arguments
```
${dir}/srcs/tmLQCD/configure \
  --prefix=$HOME/.usr \
  --enable-mpi \
  --enable-sse2 \
  --enable-sse3 \
  --with-mpidimension=4 \
  --enable-gaugecopy \
  --with-limedir=$HOME/.usr \
  --with-lapack="-L/opt/intel/mkl/lib/intel64 -Wl,--no-as-needed -lmkl_intel_ilp64 -lmkl_gnu_thread -lmkl_core -lgomp -lpthread -lm -ldl" \
  CC=mpicc CXX=mpicxx
```
The config script is fine, but when I run `make` I get a lot of error message along the lines of
```
In file included from ${dir}/srcs/tmLQCD/global.h:61:0,
                 from ${dir}/srcs/tmLQCD/operator/tm_operators.c:32:
${dir}/srcs/tmLQCD/operator/mul_one_pm_imu_sub_mul_body.c: In function ‘mul_one_pm_imu_sub_mul’:
${dir}/srcs/tmLQCD/sse.h:276:1: error: memory input 0 is not directly addressable
 __asm__ __volatile__ ("movapd %0, %%xmm0 \n\t" \
 ^
${dir}/srcs/tmLQCD/su3.h:176:1: note: in expansion of macro ‘_sse_load’
 _sse_load(s1); \
 ^
${dir}/srcs/tmLQCD/operator/mul_one_pm_imu_sub_mul_body.c:38:5: note: in expansion of macro ‘_vector_su
’
     _vector_sub(t->s0, phi1, s->s0);
     ^
${dir}/srcs/tmLQCD/sse.h:276:1: error: memory input 1 is not directly addressable
 __asm__ __volatile__ ("movapd %0, %%xmm0 \n\t" \
 ^
${dir}/srcs/tmLQCD/su3.h:176:1: note: in expansion of macro ‘_sse_load’
 _sse_load(s1); \
 ^
${dir}/srcs/tmLQCD/operator/mul_one_pm_imu_sub_mul_body.c:38:5: note: in expansion of macro ‘_vector_su
’
     _vector_sub(t->s0, phi1, s->s0);
     ^
${dir}/srcs/tmLQCD/sse.h:276:1: error: memory input 2 is not directly addressable
[...]

PR #394, opened 2017-08-01
Fix issue #393

urbach:
This is a fix to issue #393.

    The problem are the explicit SSE version of non-even/odd D_psi and
    Dsw_psi.
    The only feasible workaround at the moment is to disable SSE2|3 in
    D_psi.
    This will not affect the even/odd versions, which are still available
    in SSE2|3.
    
    In the future we'll remove these macros completely.

### DIFF
--- a/operator/D_psi.c
+++ b/operator/D_psi.c
@@ -49,7 +49,7 @@
 #include "boundary.h"
 #include "gamma.h"
 #include "linalg_eo.h"
-#ifdef MPI
+#ifdef TM_USE_MPI
 # include "xchange/xchange.h"
 #endif
 #include "update_backward_gauge.h"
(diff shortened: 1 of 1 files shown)


kostrzewa:
```MPI``` should be ```TM_USE_MPI``` and ```OMP``` should be ```TM_USE_OMP```

urbach:
merci, should be fixed now

> ```MPI``` should be ```TM_USE_MPI``` and ```OMP``` should be ```TM_USE_OMP```
> 
> -- 
> You are receiving this because you authored the thread.
> https://github.com/etmc/tmLQCD/pull/394#issuecomment-319652359
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
Now would probably also be a good time to expand which functionality is covered by travis...

martin-ueding:
I suppose you want to build with different sets of  `--enable-sse2` and `--enable-sse3`? One can use the Travis CI build matrix and supply various sets of environment variables. The builders support AVX but not AVX2, so SSE2 should be fine but SSE3 will likely not be supported.

kostrzewa:
This looks fine to me, meging.

PR #398, opened 2017-09-01
Invert the correct flavour in invert_quda_direct 

kostrzewa:
(this version of the code has never been used in production). Do not reorder the input spinor after the inversion (not sure why this was ever done). In addition, use threads in reordering routines.


kostrzewa:
ping

Issue #400, opened 2017-10-18
DDalphaAMG threading issue


kostrzewa:
@sbacchio @urbach @Finkenrath 

I think I found the reason why the threading doesn't work right now... in DDalphaAMG interface:

```
#ifdef OMP
  if(mg_omp_num_threads<=0)
      mg_init.number_openmp_threads=omp_num_threads;
  else
      mg_init.number_openmp_threads=mg_omp_num_threads;
#else
  mg_init.number_openmp_threads=1;
#endif
```

should be
```
#ifdef TM_USE_OMP
  if(mg_omp_num_threads<=0)
      mg_init.number_openmp_threads=omp_num_threads;
  else
      mg_init.number_openmp_threads=mg_omp_num_threads;
#else
  mg_init.number_openmp_threads=1;
#endif
```

kostrzewa:
@chelmes: we will need to redo the benchmarks on Marconi A2!

kostrzewa:
Okay, so threading now works, but the scaling is a little poor.

1 thread:
```
# Time for cloverdetratio3light monomial derivative: 7.078156e+00 s
```

7 threads:
```
# Time for cloverdetratio3light monomial derivative: 4.506283e+00 s
```

But that's something at least. This should allow us to run on much larger machine partitions on SuperMUC and similar machines.

kostrzewa:
The reason I didn't notice the bug is because DDalphaAMG (or the interface) suppresses the log output

"running with N openmp threads per core"

if

 ```
mg_init.number_openmp_threads==1;
```

@sbacchio : What's your feelling about the observed speed-up (7 to 4.5 seconds), can this be improved?

sbacchio:
Ok, good to know! :)
Yes, yesterday when you explained it, it was clear was a kind of bug.
I knew that the openMP implementation was not as good as MPI but at least it was doing something.
I'm sorry I used the wrong flag...

So regarding the timings I guess that's for KNL? or with a high number of MPI processes?
What I remember is that openMP was working fine if used in replacement of MPI tasks. Instead over the maximum number of MPI tasks was quite poor.

kostrzewa:
> So regarding the timings I guess that's for KNL? or with a high number of MPI processes?
What I remember is that openMP was working fine if used in replacement of MPI tasks. Instead over the maximum number of MPI tasks was quite poor.

[...]

PR #402, opened 2017-11-02
Fix GH#400, DDalphaAMG interface should use correct TM_USE_OMP prepro…

kostrzewa:
…cessor macro

### DIFF
--- a/DDalphaAMG_interface.c
+++ b/DDalphaAMG_interface.c
@@ -435,7 +435,7 @@ void MG_init()
   mg_init.theta[3] = X1;
 
   mg_init.number_of_levels=mg_lvl;
-#ifdef OMP
+#ifdef TM_USE_OMP
   if(mg_omp_num_threads<=0)
       mg_init.number_openmp_threads=omp_num_threads;
   else


kostrzewa:
This should fix #400 . One needs to be careful to specify ```MGOmpNumThreads``` explicitly, however, because it is often better to have fewer threads in DDalphaAMG than in tmLQCD/QPhiX. Right now, it is not clear to me if this doesn't have the side-effect of reducing the thread count everywhere. In principle one should thus call ```omp_set_num_threads() ``` in tmLQCD upon leaving the DDalphaAMG context.

kostrzewa:
I will merge this now and we'll worry about the consequences later. In combination in QPhiX, there is no problem because QPhiX sets the number of threads each time a geometry is created (which is at each solve in our case, causing some limited overhead). Similarly, DDalphaAMG sets the number of threads each time it's called, so we can safely specify different numbers of threads.

What I'm not sure about, of course, are the other parts of tmLQCD. There is probably one side-case which could even create real problems. This is related to the reduction arrays which depend on the number of threads. As a result, ```MGOmpNumThreads``` should never be chosen larger than ```OMPNumThreads```.

One way of resolving this relatively easily would be to save the current number of threads in the DDalphaAMG interface (I'm referring to the interface code within tmLQCD), start the DDalphaAMG solve and then reset the number of threads upon returning.

@sbacchio, @Finkenrath : could you mark that as a TODO in the DDalphaAMG interface code?

PR #403, opened 2017-11-08
Qphix devel

kostrzewa:
Finally, this is the full QPhiX interface, ready to be merged into master.


kostrzewa:
I might remove the QPhiX operator base classes, since these are essentially duplicated in QPhiX now and increase the maintenance burden. Before the final merge, I need to fix some things up in the qphix/devel branch, so I can run a final test.

kostrzewa:
It's "go" time. As soon as https://github.com/JeffersonLab/qphix/pull/112 has landed, this branch of tmLQCD will compile out of the box with the QPhiX devel branch. We've lost the qphix Dslash test in the process and it should be recovered at some point (at the latest when we plan to get the QPhiX interface covered by Travis), but the code should be usable in production now, modulo the tests that @sunpho84 is still performing.

martin-ueding:
The pull request on the QPhiX side has been merged, so you can go ahead here.

kostrzewa:
Okay, I'm just going to go ahead, otherwise we will never merge this...

Issue #405, opened 2017-11-17
QPhiX defaults to relative residual, need to adjust interface


kostrzewa:
Since the MG mods have been merged into QPhiX, all solvers have been equipped with an additional flag to switch between absolute and relative residual, the latter being the default. In the QPhiX interface, we currently calculate the appropriate absolute residual to fulfil whatever criterion has been set in the tmLQCD input file, but since the changes in QPhiX are new, we don't explicitly set the absolute residual flag.

As a first step, we should explicitly pass the absolute residual flag in all QPhiX solver calls or, if more time is available to implement the changes, we should switch to let QPhiX handle this directly as it is now supported.

kostrzewa:
Okay, it seems that the merges and developments on QPhiX have introduced a bug somewhere, which causes NaNs with all single flavour monomials. At least the NaN appears from iteration one and the initial |r|^2 is zero, so it's pretty clear that the relative residual mechanism doesn't work properly. The first step will be to explicitly do absolute residual.

kostrzewa:
Okay, the problem does not seem to go away by doing that... damn.

kostrzewa:
Ah, maybe the problem is in my test. tmLQCD itself also nans...

kostrzewa:
Okay, luckily the issue was simply that I had no timescale defined for the monomials under test which meant that there were no valid pseudofermion fields..

kostrzewa:
Okay, I think I've worked out all the kinks. So, QPhiX was doing RELATIVE by default and the interface was built with that in mind already, which is god news. In #407 I make this more explicit. This can be closed for now.

PR #407, opened 2017-11-19
qphix_interface adjust for new residual flag in QPhiX

kostrzewa:
call all QPhiX solvers which support it with the QPhiX::RELATIVE flag, thus making explicit how we make the tmLQCD and QPhiX residuals agree for whatever is given in the input file


kostrzewa:
Takes care of #405

PR #411, opened 2017-12-28
Final QUDA-MG interface

kostrzewa:
This is mostly done. The main remaining issue is that of enabling size 3 aggregrates, which seem to be currently unsupported by a default QUDA installation due to lack of the necessary instantiations (at least for the lattice size that was tested). Having support for these would probably be useful, however.

The default parameters selected here seem to work rather well as is, giving speedups of 3 or so over BiCGstab for Wilson clover fermions at around 270 MeV and speedups of well over 6 for twisted clover fermions at around 240 MeV. These speedups are likely about twice as large around the physical light quark mass.

These speedups are calculated against the minimum number of cards that a given problem fits on. While the CG and BiCGstab solves fit on a single P100, four cards are required for the corresponding MG solve. These are thus "worst-case" speedups, where the simple solvers are running in the most efficient way possible.


kostrzewa:
- [ ] Check whether compilation without QUDA works as expected (travis should do this)
- [ ] See if parameter scans may reveal better default parameters.
- [ ] Add support for setting number of vectors, currently this defaults to 24.

kostrzewa:
Just need to update the docs and this can be merged I think.

kostrzewa:
Since the Schwarz-preconditioner is an experimental feature, I would like to postpone its inclusion in the interface to a later date. From my side, this should now be complete and functional.

kostrzewa:
@urbach I would be happy to get this merged. It would be great if you could take a look to see if you spot anything untoward...

kostrzewa:
So, I've resolved https://github.com/etmc/tmLQCD/issues/420 mostly. For measurements, everything runs with gauge persistence mode and ```nstore``` is used to make sure that the currently loaded gauge configuration on host memory and the gauge configuration in device memory, as well as all depdendent fields, match. This makes doing many inversions on the same gauge background using the default ```invert_quda``` code path with residual checks quite a bit faster because it removes superfluous host-device transfers. For the MG, this change is essential because the setup is now also persistent and will only be recomputed when necessary.

Note that our travis is broken because some package or other changed on travis...

kostrzewa:
This also paves the way, potentially, for using QUDA solvers in the HMC, but we need to write some wrappers for passing eo-ordered sources to the QUDA solver, not sure how easy / hard this will be.

The "gauge id" tracking can then be used, perhaps augmented to a ```double``` ```(nstore + tau*curr_step/N_steps``` or something like that to a) update the fields as we move along the trajectory, b) run the MG setup update when a given number of integration sub steps has been performed. (much like we do in DDalphaAMG)

kostrzewa:
It would be great if we could merge this...

invert_clover_eo.c
```diff
@@ -79,7 +79,7 @@ int invert_clover_eo(spinor * const Even_new, spinor * const Odd_new,
       return invert_eo_quda(Even_new, Odd_new, Even, Odd,
                             precision, max_iter,
                             solver_flag, rel_prec,
-                            1, solver_params,
+                            even_odd_flag, solver_params,
```
urbach: I guess you checked that the correct mapping to 0,1 is used here

kostrzewa: I don't quite understand, ```even_odd_flag``` specifies whether the solver should use e/o preconditioning or not. Further down the line, when this is set to ```1```, this sets the QUDA solve type.

kostrzewa:
Note that there is still a major caveat here... Right now, the the device creation of the clover field and the MG setup are performed for a particular csw, kappa and mu and then re-used unless the gauge id counter has changed. If one now were to do inversions with different mu values, one would run into trouble, I believe. In practice we use ```QUDA_DYNAMIC_CLOVER```, which I think will result in the correct thing being done regardless in terms of getting the right solution, but I need to actually try and see what happens when ```AddDownPropagator``` is passed...

kostrzewa:
Indeed, this does not work. Let's see what answers emerge from https://github.com/lattice/quda/issues/709

kostrzewa:
Okay, that last issue is now also resolved. Of course, in practice one would calculate up and down separately, but at least the program flow with ```AddDownPropagator``` works as expected also with the MG solver.

urbach:
> kostrzewa commented on this pull request.
> 
> 
> 
> > @@ -79,7 +79,7 @@ int invert_clover_eo(spinor * const Even_new, spinor * const Odd_new,
>        return invert_eo_quda(Even_new, Odd_new, Even, Odd,
>                              precision, max_iter,
>                              solver_flag, rel_prec,
> -                            1, solver_params,
> +                            even_odd_flag, solver_params,
> 
> I don't quite understand, ```even_odd_flag``` specifies whether the solver should use e/o preconditioning or not. Further down the line, when this is set to ```1```, this sets the QUDA solve type.

sorry, was a confusion from my side...

kostrzewa:
One more modification required: for some reason I missed that the setup needs to be tossed out and redone completely (rather than updated) when a new configuration is loaded. I thought I had that covered, but apparently not. It's an easy fix, however.

kostrzewa:
Okay, so I've introduced some changes which should cover the situations that one encounters in practice. Technically, one should probably also introduce thresholds for the various operator parameters, i.e. if:

* c_sw changes significantly, one might have to redo the setup rather than just update the coarse operators
* similarly for kappa, although here one should be careful because the relevant quantity is the subtracted mass (increasing that should be no problem while decreasing it will affect how well the setup preconditions the solve)
* for mu the same issues apply as for kappa: increasing is fine, decreasing can be problematic

We'll touch on those issues if and when we get there, however!

kostrzewa:
Okay, I've tested the ```MGResetSetupThreshold``` input parameter and it works as expected, causing the setup to be redone completely if the ```gauge_id``` of the current setup exceeds the ```gauge_id``` of the gauge field by more than this threshold. This should also prepare us for using QUDA-MG in the HMC down the line, at least from the point of view of "overall structure".

However, there is still something going wrong at the QUDA end of the chain with some illegal memory access occurring. Might require a more recent QUDA version for testing...

In any case, the interface as it is can be used when only a single gauge field is done in a given job.

kostrzewa:
DDalphaAMG supports twisted boundary conditions natively, there's no need to do anything when switching from one twist to another.

kostrzewa:
For QUDA-MG, the setup will at the very least need to be updated and I'm not sure whether it might not even need to be reset. Currently, if we use twisted boundary conditions, the phases are multiplied into the gaugefield when using QUDA.

sunpho84:
Really? I know that in DD the twist can be specified, and the phase gets incorporated in the configuration on the coarse level, but does that have an impact on the setup of the fine operator?
I think that the spectrum of Dirac operator, and its eigenvectors, change dramatically when changing b.c., is that taken into account in the setup? To the least, I guess some phasing of the null vector is due, but the impact of twisting is difficult to imagine... I understood from @sbacchio @sbacchio @sbacchio that it was not... Did I miss something?

sbacchio:
Yes, also in DDalphaAMG the shift is added directly to the gauge-links, but here it is done in the interface on the DDalphaAMG side and not on the tmLQCD side.
It's easy to try if an update of the setup is enough, but I agree with @sunpho84 that changing the twist in the gauge-field will considerably affect the eigenmodes; and specially at the physical point is very important to construct the MG setup correctly.

kostrzewa:
@sunpho84 @sbacchio Hmm, I think I was remembering incorrectly. Yes, in the vector form factor calculation, I performed all inversions for a given theta in one job and then even restarted tmLQCD because it of course doesn't support changing the twist angle mid-execution. You are right.

kostrzewa:
I would like to merge this ASAP since we will need it for the eta-eta project tests which will take place on QBIG soon (@marcus-petschlies @blons will require these changes)

urbach:
for me this is fine to be merged in. Just checks are failing...

kostrzewa:
Fixed in https://github.com/etmc/tmLQCD/pull/431

urbach:
okay, if there is no protest I'll merge this in

kostrzewa:
To be safe, you can merge 431 and I'll update this pull such as to include those changes, then the checks will go through (or compilation will fail, if I messed up)

urbach:
merged already...
--
Carsten Urbach
e-mail: curbach@gmx.de
        urbach@hiskp.uni-bonn.de
Fon   : +49 (0)228 73 2379
skype : carsten.urbach

kostrzewa:
Good thing we checked, there is indeed a build problem..

kostrzewa:
Not my fault though, apparently. gcc is having trouble with the ```isnan``` macro in ```io/utils.h```

urbach:
hmm, so what do we do...?

kostrzewa:
I'll try to fix it. I can't reproduce it locally.

kostrzewa:
I think I know what the problem is, there was a commented out line in ```travis-ci.sh``` which broke a multi-line command

kostrzewa:
This meant that tmLQCD was being compiled without the --std=c99 flag

kostrzewa:
In c99, a mulit-purpose isnan was introduced, such that the hand-written one was never ever called anywhere else. Should I remove the hand-written one? This seems to be a possible source of odd build errors.

martin-ueding:
Yes, bash does not handle that well, had hit me in the chroma installation script multiple times 😒

PR #414, opened 2018-01-10
gradflow topological charge measurement

kostrzewa:
## TODO
- [x] Our OpenMP accumulators assume that only one quantity is accumulated, this needs to be generalised.
- [x] It must be ensured that Q is correctly normalised.
- [x] Output of Q must be enabled
- [x] 1-to-1 cross-check of energy density measurement on realistic lattice to make sure that I didn't break anything

### DIFF
--- a/default_input_values.h
+++ b/default_input_values.h
@@ -200,4 +200,8 @@
 #define _default_subprocess_flag 0
 #define _default_lowmem_flag 0
 
+// gradient flow measurement step size and maximum flow time
+#define _default_gf_eps 0.01
+#define _default_gf_tmax 9.99
+
 #endif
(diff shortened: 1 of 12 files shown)


kostrzewa:
@urbach: This version contains measurements of the communication overhead at ```g_debug_level >= 4```, so we can check if it makes sense to implement two-row gauge compression for communication.

kostrzewa:
Okay, it seems that just one more check is required and we'll have a working topo charge measurement.

kostrzewa:
The code has now been validated on three gauge configurations against @uwenger's (thanks!) and only round-off differences were found, which likely arise due to our use of Kahan sums in the process and thread-local reductions.

PR #418, opened 2018-02-25
tmLQCD_invert_init should return if the input file cannot be found

kostrzewa:
expose global.h and struct_accessors.h through tmLQCD.h, return error state in lib_wrapper if invert.input cannot be found

### DIFF
--- a/include/tmLQCD.h
+++ b/include/tmLQCD.h
@@ -29,9 +29,6 @@
 
 #include "config.h"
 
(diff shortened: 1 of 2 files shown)


kostrzewa:
Actually, I'm not sure about global.h anymore...

kostrzewa:
I removed the inclusion of ```struct_accessors.h``` and ```global.h``` because it leads to problems in C++ applications...

PR #422, opened 2018-07-10
DO NOT MERGE: this is a comparison PR between Finkenrath/DDalphaAMG_nd  and etmc/master

kostrzewa:


expo.c
```diff
@@ -52,48 +52,132 @@
...
+  vt.c02 =      p1 * v.c02 + p2 * v2.c02;
+  vt.c10 =      p1 * v.c10 + p2 * v2.c10;
+  vt.c11 = p0 + p1 * v.c11 + p2 * v2.c11;
+  vt.c12 =      p1 * v.c12 + p2 * v2.c12;
+  vt.c20 =      p1 * v.c20 + p2 * v2.c20;
+  vt.c21 =      p1 * v.c21 + p2 * v2.c21;
+  vt.c22 = p0 + p1 * v.c22 + p2 * v2.c22;
+  
+  for(m=0;m<mm;m++) {
+    _su3_times_su3(v2,vt,vt);
```
kostrzewa: assuming that all coefficients are as they were before (but now computed algorithmically), why does vt get squared (potentially) multiple times here?

kostrzewa: @Finkenrath

hmc_tm.c
```diff
@@ -496,12 +498,22 @@ int main(int argc,char *argv[]) {
     }
 
     /* online measurements */
+#ifdef DDalphaAMG
+    // When the configuration is rejected, we have to update it in the MG and redo the setup.
```
kostrzewa: @Finkenrath @urbach @sbacchio 

Wait a second, why was this never merged into tmLQCD/master? Does this mean that all simulations done with tmLQCD/master have incorrect online measurements?! Or was this done internally in the DDalphaAMG interface previously and only later placed here?

kostrzewa: Actually, I don't understand. Looking at my local copy of tmLQCD/master, these lines are present... I guess this comparison here is flawed because perhaps there were some rebases or something like that?

PR #423, opened 2018-07-11
merge DDalphaAMG_nd branch into etmc/tmLQCD/master

kostrzewa:


kostrzewa:
@urbach @Finkenrath @sbacchio 

okay, so this is the correct comparison between etmc/tmLQCD [master] and Finkenrath/tmLQCD [DDalphaAMG_nd]

How should we proceed?

invert.c
```diff
@@ -179,6 +180,9 @@ int main(int argc, char *argv[])
   j = init_gauge_field(VOLUMEPLUSRAND, 0);
   j += init_gauge_field_32(VOLUMEPLUSRAND, 0);  
 #endif
+  if(restoresu3_flag) {
```
kostrzewa: @Finkenrath @sbacchio 

There is no default set for this anywhere, if I'm not mistaken. This should be done in read_input as for the other global input paramters.

monomial/ndrat_monomial.c
```diff
@@ -229,28 +230,16 @@ void ndrat_heatbath(const int id, hamiltonian_field_t * const hf) {
   }
   mnl->solver_params.sdim = VOLUME/2;
   mnl->solver_params.rel_prec = g_relative_precision_flag;
-  mnl->iter0 = solve_mms_nd(g_chi_up_spinor_field, g_chi_dn_spinor_field,
-                            mnl->pf, mnl->pf2, &(mnl->solver_params) );
+  mnl->iter0 = solve_mms_nd_plus(g_chi_up_spinor_field, g_chi_dn_spinor_field,
```
kostrzewa: Does this change something when other solvers are used?

kostrzewa:
There are so many interconnected changes here it's very difficult to review...

update_momenta_fg.c
```diff
@@ -128,6 +129,7 @@ void update_momenta_fg(int * mnllist, double step, const int no,
       exposu3(&w,&deriv);
       restoresu3(&v,&w);
       _su3_times_su3(w, v, *z);
+      restoresu3(&v,&w);
       _su3_assign(*z, w);
```
kostrzewa: @Finkenrath 
The final ```restoresu3``` doesn't do anything here. One line below, ```w``` is assigned to ```z```, while the "restored" su3 matrix is in ```v```.

Finkenrath: Yes, one of the calls of restoresu3 canbe deleted. In case of openQCD the functions 
exposu3(&w,&deriv); _su3_times_su3(w, v, *z);
are contained in one function and the restoration of SU3 is done after the multiplication. I would prefer to delete the first call and have the second, then we are sure that the global gaugefield is in SU3.

kostrzewa: The point was that the second one is actually not doing anything as it is right now.

kostrzewa: > are contained in one function and the restoration of SU3 is done after the multiplication

I would expect that to be overkill. This would mean that whenever one computes a plaquette or a staple, multiple restoresu3 are called.

kostrzewa: Ah, I see what you mean now. In OpenQCD, the restoration is done right after the exponentiated momenta are multiplied to the gauge field. We restore the su3'ness of the gauge field once per trajectory instead.

As for the exponentiated momenta, @urbach, do you think restoring the result of the exponentiation to su3 is important at this stage? We do it in the current implementation.

monomial/ndrat_monomial.c
sbacchio: What do you mean? Probably I didn't understand the question.
For clarity sake, now we are calling a function which should return the solution of the plus and not the squared operator. This change is combined with the lines removed after where the operator with negative sign is applied (Qsw_tau1_sub_const_ndpsi). In case of MG the plus operator is directly inverted. In case of CG the squared operator is inverted and then the operator with negative sign is applied.

kostrzewa: I see, the operator was applied below before, now it is done in the solver driver. I missed the lines below on the left (240 - 253).

read_input.l
```diff
@@ -800,6 +804,17 @@ static inline double fltlist_next_token(int * const list_end){
     mg_omp_num_threads=a;
     if(myverbose) printf("  MG_omp_num_threads set to %d line %d operator %d\n", mg_omp_num_threads, line_of_file, current_operator);
   }
+  {SPC}*MGNumberOfShifts{EQL}{DIGIT}+ {
+    sscanf(yytext, " %[a-zA-Z] = %d", name, &a);
+    mg_no_shifts=a;
+    mg_mms_mass=0;
```
kostrzewa: @sbacchio 
This cannot be done here. If the order in the input file is reversed (MGMMSMass is specified before MGNumberOfShifts), then mg_mms_mass will always be reset to 0.

Default parameters should always be set in the main section of the input file parser, which is run before anything else.

update_momenta_fg.c
kostrzewa: @Finkenrath The second ```restoresu3``` in line 132 still doesn't do anything. In line 133, you should assign ```v``` instead of ```w``` if what you wrote there was what you intended...

read_input.l
sbacchio: The idea is that either MGMMSMass or MGNumberOfShifts should be used. Both cannot be used. As it is implemented is a consistent way to use either one of the other and in case both are called in the input file; then the last will be used. I'm sorry but I did it, I didn't find a better way to implement is giving both options. For sure I need to set mg_mms_mass=0 when MGNumberOfShifts is used in order to make the latter work.

kostrzewa: Okay, that's fine then.

kostrzewa: I'll add a comment to mark the intent.

kostrzewa: Should MGMMSMass also modify the input parameter for the number of shifts, such that the two really are exclusive?

sbacchio: Accordingly to the current implementation no, because the check on MGMMSMass has priority to the check on the number of shifts... but for clarity sake would be worth to add in MGMMSMass a line that set `mg_no_shifts=0` so the intent is clear. Since you add the comment, push also this change :)

kostrzewa: Fine. Do you think you'd have the time to extend the documentation for the DDalphaAMG interface to include all these new input parameters?

kostrzewa: ping

update_momenta_fg.c
kostrzewa: ping

read_input.l
```diff
@@ -1883,14 +1931,14 @@ static inline double fltlist_next_token(int * const list_end){
   }
 }
 
-<RATMONOMIAL,CLRATMONOMIAL,RATCORMONOMIAL,CLRATCORMONOMIAL>{
+<RATMONOMIAL,RATCORMONOMIAL,CLRATMONOMIAL,CLRATCORMONOMIAL>{
   {SPC}*Solver{EQL} {
-    solver_caller=YY_START;
-    BEGIN(RATMSOLVER);
+   solver_caller=YY_START;
+   BEGIN(MSOLVER);
```
kostrzewa: Is this a safe modification? There is a reason why there was a RATMSOLVER...

read_input.l
```diff
@@ -1883,14 +1931,14 @@ static inline double fltlist_next_token(int * const list_end){
...
   }
 }
 
-<DETMONOMIAL,POLYMONOMIAL,CLDETMONOMIAL,CLDETRATMONOMIAL,CLDETRATRWMONOMIAL>{
+<DETMONOMIAL,POLYMONOMIAL,CLDETMONOMIAL,CLDETRATMONOMIAL,CLDETRATRWMONOMIAL,RATMONOMIAL,RATCORMONOMIAL,CLRATMONOMIAL,CLRATCORMONOMIAL,NDDETRATMONOMIAL,NDCLDETRATMONOMIAL>{
```
kostrzewa: Why these changes? There are parameters here which should not be avaialable for the RATMONOMIAL types...

solver/cg_her_nd.c
```diff
@@ -156,10 +154,15 @@ int cg_her_nd(spinor * const P_up,spinor * P_dn, spinor * const Q_up, spinor * c
...
+
+  etime = gettime();
+  if(g_debug_level > 0 && g_proc_id == 0) {
+    printf("# CG: iter: %d eps_sq: %1.4e t/s: %1.4e\n", iteration, eps_sq, etime-atime); 
+  }
   
   finalize_solver(up_field, nr_sf);
   finalize_solver(dn_field, nr_sf);
-  return(-1);
+  return(iteration+1);
```
kostrzewa: convergence failure is given by the return of ```-1```, why this change of the logic?

smearing/hex_stout_exclude_none.c
```diff
@@ -3,6 +3,7 @@
 void stout_exclude_none(su3_tuple *buff_out, double const coeff, su3_tuple **staples, su3_tuple *buff_in)
 {
   static su3 tmp;
+  if (g_exposu3_no_c == 0) init_exposu3();
```
kostrzewa: This should be called at program start and not at various places in the code-base.

solver/rg_mixed_cg_her_nd.c
```diff
@@ -248,17 +248,27 @@ int rg_mixed_cg_her_nd(spinor * const P_up, spinor * const P_dn, spinor * const
...
-  zero_spinor_field(P_up,N); zero_spinor_field(P_dn,N);
-  assign(phigh_up,Q_up,N); assign(phigh_dn,Q_dn,N);
-  assign(rhigh_up,Q_up,N); assign(rhigh_dn,Q_dn,N);
-  
-  rho_dp = ( square_norm(rhigh_up,N,1) + square_norm(rhigh_dn,N,1) );
+
+  guesssquarenorm = square_norm(P_up, N, 1);
+  guesssquarenorm += square_norm(P_dn, N, 1);
+
+  if(guesssquarenorm == 0) {
```
kostrzewa: This comparison is not safe, nor do you know what is in P_up and P_dn...

invert.c
kostrzewa: ping

read_input.l
kostrzewa: I meant the ND[...]RATMONOMIAL types

read_input.l
```diff
@@ -1757,15 +1793,15 @@ static inline double fltlist_next_token(int * const list_end){
...
   {SPC}*Kappa{EQL}{FLT} {
     sscanf(yytext, " %[a-zA-Z] = %lf", name, &c);
     mnl->kappa = c;
     if(myverbose) printf("  Kappa set to %f line %d monomial %d\n", c, line_of_file, current_monomial);
   }
 }
 
-<DETMONOMIAL,POLYMONOMIAL,CLDETRATRWMONOMIAL>{
+<DETMONOMIAL,POLYMONOMIAL,CLDETRATRWMONOMIAL,NDDETRATMONOMIAL,NDCLDETRATMONOMIAL>{
   {SPC}*2KappaMu2{EQL}{FLT} {
```
kostrzewa: I don't think either NDDETRATMONOMIAL nor NDCLDETRATMONOMIAL should have a ```2KappaMu2``` parameter...

read_input.l
kostrzewa: or are shift-by shift solves supported now for all of these monomials using all solvers that can be chosen via ```MSOLVER```?

read_input.l
```diff
@@ -2106,15 +2154,25 @@ static inline double fltlist_next_token(int * const list_end){
     mnl->solver = BICGSTAB;
     BEGIN(solver_caller);
   }
+  cgmms {
```
kostrzewa: CGMMS is not a solver which can be used for DETMONOMIAL or CLDETMONOMIAL.

read_input.l
```diff
@@ -2124,6 +2182,11 @@ static inline double fltlist_next_token(int * const list_end){
     mnl->solver = CGMMSND;
     BEGIN(solver_caller);
   }
+  rgmixedcg {
```
kostrzewa: does this now expose the shift-by-shift solve using rgmixedcg?

kostrzewa:
Quite a few things left to check / change / fix. Also, there might be conflicts with @gbergner's reweighting branch https://github.com/etmc/tmLQCD/pull/349

invert.c
sbacchio: This was just a check for computing the violation of the special unitary of the links due to the bug in restore su3.
We applied reweighed the action of some old ensembles for testing that difference in the energy wasn't significant.
All the tests have been passed and this change can be removed if you wish.
The full change is visible in commit 867b24afd15a3e199b1f21d84f035c1d7d3d1b39
https://github.com/etmc/tmLQCD/pull/423/commits/867b24afd15a3e199b1f21d84f035c1d7d3d1b39

update_momenta_fg.c
sbacchio: I'm going to push the changes on DDalphaAMG_nd

read_input.l
sbacchio: Done, it will be available in the coming push

read_input.l
sbacchio: Exactly.

read_input.l
sbacchio: Correct, going to fix it.

invert.c
kostrzewa: It's okay to leave in, but then a default needs to be set for the parameter. Since it's just a check, perhaps it should be removed?

update_momenta_fg.c
kostrzewa: No, please fork my branch and push there

kostrzewa: Otherwise I need to merge all over again... or just merge the current master, but then we're doing work twice..

read_input.l
kostrzewa: Awesome, thanks.

update_momenta_fg.c
sbacchio: Of course I've merged your branch.

read_input.l
sbacchio: Fine I will restore RATMSOLVER. By the way monomial solve should do the work since it doesn't have in the list of single shift CGMMS.

read_input.l
sbacchio: Yes

read_input.l
```diff
@@ -2908,15 +2973,23 @@ static inline double fltlist_next_token(int * const list_end){
...
   reweighting_flag = 1;
-  if(myverbose!=0) fprintf(stderr, "Compute reweighting factor\n");
+  if(myverbose!=0) printf("Compute reweighting factor\n");
 }
 <REWEIGH>no {
   reweighting_flag = 0;
-  if(myverbose!=0) fprintf(stderr, "Do not compute reweighting factor\n");
+  if(myverbose!=0) printf("Do not compute reweighting factor\n");
+}
+<RESTORESU3>yes {
```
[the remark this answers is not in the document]
sbacchio: Accordingly to the following, external variables are always initialized to zero. Anyway I've added the default initialization.
https://stackoverflow.com/questions/6212921/is-un-initialized-integer-always-default-to-0-in-c

solver/cg_her_nd.c
sbacchio: No reasons. I've fixed it. But this arise another question... why return a -1 if then there is no check somewhere else? specially in solve_monomial?
In case of HMC one wouldn't notice that there has been a convergence failure since the number of iterations are summed in a total and nowhere is checked if the values is -1.

read_input.l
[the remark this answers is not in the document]
kostrzewa: Okay, good. I'd rather not depend on the standard too much though. Better safe than sorry if we have to rely on some shoddy compiler at some point. XLC comes to mind...

Anyway, about the SU(3) restoration stuff: This currently introduces a lot of cruft into invert and for the sake of consistence, the same cruft should be available in offline_measurement. It would be good to spin off the calls made in invert.c for this into functions, such that they can easily be called elsewhere too. I know, cleaning up part of invert.c is like mopping the floor in a sewer, but maybe we can inch closer to finally reorganizing the code-base in such a way that one is not inclined to remove one's eyeballs upon gazing at the mess... @gbergner and I have been discussing to what extent it will be possible with some not too extreme effort to clean things up in the next few weeks.

read_input.l
kostrzewa: > By the way monomial solve should do the work since it doesn't have in the list of single shift CGMMS.

I'm not sure what you mean

solver/cg_her_nd.c
kostrzewa: Yeah, I know, it's insane.

kostrzewa: But having the return value at least gives us the chance of doing this now in the various solver drivers. It's on the todo list to finally introduce checks for this...

We've actually had to repeat HMC runs because of this insanity.

solver/rg_mixed_cg_her_nd.c
sbacchio: I added a flag use_inital_guess in solver params. Its support should be extended... At the moment should be fine.

read_input.l
sbacchio: yeah

sbacchio:
Pushed, I think now you can pull

monomial/cloverdetratio_rwmonomial.c
```diff
@@ -62,6 +79,21 @@ double cloverdetratio_rwacc(const int id, hamiltonian_field_t * const hf) {
...
+        v=&(hf->gaugefield[ix][mu]);
+        restoresu3_in_place(v);
+      }
+    }
+#ifdef TM_USE_MPI
+    xchange_gauge(hf->gaugefield);
+#endif
+    mg_update_gauge = 1;
+  }
+
```
kostrzewa: shouldn't the old gauge field be restored here?

read_input.l
sbacchio: I meant that if someone tries use CGMMS as a solver for DETMONOMIAL, then solve_degenerate in solve_monomial will return error because CGMMS is not in the list of solver.
Anyway now it's removed, I've just forgot to do it before.. I've pushed again.

reweighting_factor.c
```diff
@@ -54,24 +54,32 @@ void reweighting_factor(const int N, const int nstore) {
     mnl = &monomial_list[j];
```
kostrzewa: @gbergner: I would like to pull this in as soon as possible, does this collide with your work?

linalg/square_and_minmax.c
```diff
@@ -0,0 +1,424 @@
...
+# include <omp.h>
+# include "global.h"
+#endif
+#include <complex.h>
+#include "su3.h"
+#include "su3adj.h"
+#include "su3spinor.h"
+#include "square_and_minmax.h"
+
+void square_and_minmax(double * const sum, double * const min, double * const max, const spinor * const P, const int N)
```
kostrzewa: @sbacchio are these routines used? They are not threaded...

kostrzewa: It's seems like they are not used anywhere. Should I remove them? Otherwise they should be threaded. Note that I've implemented some comfortable wrappers for doing multi-threaded Kahan sums.

kostrzewa: See ```kahan_summation.h``` and their use in ```meas/measure_clover_field_strength_observables.c```

monomial/cloverdetratio_rwmonomial.c
```diff
@@ -45,14 +45,31 @@
...
+#include "DDalphaAMG_interface.h"
 
 double cloverdetratio_rwacc(const int id, hamiltonian_field_t * const hf) {
   monomial * mnl = &monomial_list[id];
   int save_sloppy = g_sloppy_precision_flag;
   double atime, etime;
   atime = gettime();
 
+  if (restoresu3_flag) {
+    for(int ix=0;ix<VOLUME;ix++) {
```
kostrzewa: threading missing

monomial/ndrat_monomial.c
```diff
@@ -57,7 +58,7 @@ void nd_set_global_parameter(monomial * const mnl) {
   boundary(g_kappa);
   phmc_cheb_evmin = mnl->EVMin;
   phmc_invmaxev = mnl->EVMaxInv;
-  phmc_cheb_evmax = 1.;
+  phmc_cheb_evmax = mnl->EVMax;
```
kostrzewa: did you check that this change does not break the QPhiX multi-shift solve?

monomial/ndrat_monomial.c
```diff
@@ -322,26 +311,40 @@ double ndrat_acc(const int id, hamiltonian_field_t * const hf) {
...
+  int scale = 0;
 
-  mnl->EVMin = mnl->StildeMin / mnl->StildeMax;
-  mnl->EVMax = 1.;
-  mnl->EVMaxInv = 1./(sqrt(mnl->StildeMax));
+  if(mnl->type == RAT || mnl->type == CLOVERRAT ||
+     mnl->type == RATCOR || mnl->type == CLOVERRATCOR) 
+    scale = 1;
+
+  if(scale) {
```
kostrzewa: @sbacchio @urbach Okay, I've checked this now and it seems that it works correctly, also when QPhiX is used.

First line tmLQCD only, second line tmLQCD+QPhiX for sample-hmc-rat.input on a 8^3x4 lattice. I have acceptance on the 4^4 lattice, but that I can't run with QPhiX.

```
00000000 0.124045033778 127924.456342349527 0.000000e+00 220 3592 366 3384 583 0 0 5.771806e+01 5.114121e-02
00000000 0.124045033778 127924.455873184837 0.000000e+00 220 3592 363 3334 1384 0 0 3.592745e+01 5.114121e-02

```

kostrzewa:
I think there's some logic regression: running the cC211.06.80 lattice on Marconi A3, the code gets stuck in the attempt to do the NDCLOVERRATCOR heatbath. The DDalphaAMG setup works fine, but then:

```
[...]
DDalphaAMG setup ran, time 65.76 sec (41.08 % on coarse grid)
depth: 0, mu updated to 0.106586 
depth: 1, mu updated to 0.106586 
depth: 2, mu updated to 0.106586 
depth: 0, epsbar term updated to 0.107146 + ig5 0.000000 on even sites and + ig5 0.123035 on odd sites 
depth: 1, epsbar term updated to 0.107146 + ig5 0.000000 on even sites and + ig5 0.123035 on odd sites 
depth: 2, epsbar term updated to 0.107146 + ig5 0.000000 on even sites and + ig5 0.123035 on odd sites 
```

and it sits there for 20 minutes doing nothing...

kostrzewa:
Okay, PEBKAC. I had a version of DDalphaAMG before the merge of the TM2p1p1 branch. Works now. Let's see!

kostrzewa:
Okay, pending some final test runs I will merge this in in the next few hours.

kostrzewa:
Alright, did some high statistics runs especially to test the effect of 2MNFG and I can't detect any statistically significant deviations in <P> for different combinations of MPI tasks / OpenMP threads after O(70k) trajectories on a small lattice. Will merge this now.

PR #425, opened 2018-07-18
add single flavour point source generator which takes a four-element …

kostrzewa:
…array of global source coordinates (txyz) as a source location

### DIFF
--- a/include/tmLQCD.h
+++ b/include/tmLQCD.h
@@ -28,6 +28,7 @@
 #define _TMLQCD_H
 
 #include "config.h"
+#include "su3.h"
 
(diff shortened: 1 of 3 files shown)


include/tmLQCD.h
```diff
@@ -28,6 +28,7 @@
 #define _TMLQCD_H
 
 #include "config.h"
+#include "su3.h"
```
kostrzewa: not sure about this, but it's the only way to expose the function without double-casting spinors back and forth...

kostrzewa:
This seems to be a useful utility function which is currently missing in this form.

PR #428, opened 2018-07-31
using tmLQCD via libwrapper leads to inconsistent nstore

kostrzewa:
When used as a library, the gauge field can be loaded via the external application by passing an integer configuration number to ```tmLQCD_read_gauge```. Unfortunately, doing so does not set ```nstore```and thus tmLQCD may be left in an inconsistent state.


urbach:
I could merge this in. Why does travis fail?

kostrzewa:
Because of travis... They keep updating the environment which breaks builds. I don't know if one could't force the environment to remain constant somehow. It's really annoying...

PR #431, opened 2018-09-27
finally add protection from string buffer overflows in configuration …

kostrzewa:
…filenames


kostrzewa:
I would be happy to also move ```CONF_FILENAME_LENGTH``` to some global include file, such as to not duplicate this sereral times. Beyond that, the whole reading routine should probably be in its own function with consistent variable names.

PR #434, opened 2018-10-08
DO NOT MERGE various improvements necessary to make working with QUDA efficient "everywhere"

kostrzewa:


include/tmLQCD.h
```diff
@@ -28,6 +28,7 @@
 #define _TMLQCD_H
 
 #include "config.h"
+#include "su3.h"
```
kostrzewa: This is probably contenious. Exposing the ```spinor``` data type via the interface has benefits, but of course it can also lead to naming conflicts. I guess one could add further wrappers for the source generators with double pointers, dunno...

kostrzewa:
@urbach 
Okay, so here's the basic idea to track the state of the gauge field globally (and the clover and inverse clover fields, eventually). 

In ```misc_types.h``` I create some new types to hold state information. Instances of these are are added to the global scope and initialised at program start (via ```init/init_global_states```).

At various points where the gauge field is modified, the function ```update_tm_gauge_id(&gauge_state, step)``` is called which will either initialise (upon first call) or step the gauge_id counter in the ```gauge_state``` struct. There is of course some complication due to ```hf.gaugefield``` not strictly having to represent ```g_gauge_field```, but in current practice, the two are identical.

I've made some arbitrary choices:

1. When ```g_gauge_field``` is passed to ```read_gauge```, the gauge state is incremented by ```TM_GAUGE_PROPAGATE_THRESHOLD```, which is set to ```10.0``` at present.
2. SU(3) restoration causes an increment of ```TM_GAUGE_PROPAGATE_MIN```.
3. During ```update_gauge```, the integrator step is used.
4. There is also a field which specifies whether the halo has been exchanged or not and a complementary function to update this element of the struct.

I've tested this for 1000 trajectories in a single run and at that level it is quite resilient to round-off issues. The idea is that at some point in the very near future we can avoid the update and inversion of the clover field if this is consistent with the gauge field, the required clover inverse and the parameters that the latter has been created with.

At the same time, this will help to streamline the existing approach that I've taken in the QUDA interface to ensure that neither the MG setup nor the various device fields are updated / recreated too frequently. Having this well-defined state means that I simply have to compare to that, rather than having to rely on some internal tracked state (see QUDA interface for how this is done presently). This will help when we decide to implement a QUDA monomial solver interface.

Similarly, the corresponding updates in the DDalphaAMG interface can be integrated with this, thus unifying this aspect of the code.

Finally, we might also decide to switch to keeping the QPhiX operators resident and updating their respective field copies as required by testing their state against those tracked in these new global variables.

Of course, a much cleaner approach would be to wrap all fields in structs and to keep all this meta-data tightly coupled to the fields themselves. This would involve a larger number of modifications in the entire code base, but it might be worth the effort even in the short run...

I would really like to hear opinions before we proceed further.

kostrzewa:
If this kind of approach looks good, the ```g_update_gauge_copy``` can be removed in favour of adding an element to the present struct with the same effect.

kostrzewa:
On a different note I tried to make the requested MPI thread level configurable via the input file (this might be important when QUDA is used on something like PizDaint), but I'm afraid that this of course doesn't work properly. The input file is read after MPI is initialised and if I reverse the order of operations, the input file reader produces too much output...

I guess I will have to convert this to a command line argument...

kostrzewa:
At high debug levels (4 and above), this behaves as shown in the attachment below:

[2traj.txt](https://github.com/etmc/tmLQCD/files/2465309/2traj.txt)

urbach:
@kostrzewa 
to me this looks like a good way forward. It seems to be the sort of minimal way of implementing this.

kostrzewa:
Okay, I'll proceed then. Thanks!

kostrzewa:
@pittlerf Once Harald has set up multi-threaded OpenMPI, you will need to use this branch to test it on QBIG, passing the flag ```-m multiple``` to the executable.

PR #443, opened 2018-11-05
QUDA MG solver and smoother settings as well as support for tracking TBCs

kostrzewa:


kostrzewa:
@marcus-petschlies Since you have this running on QBIG, can you test if these changes don't break anything?

kostrzewa:
I would be happy to also pull in your changes to the wrapper functions here so there's no hassle.

PR #445, opened 2018-11-09
Do MPI initialisation in tmLQCD_invert_init

kostrzewa:
It seems that having tmLQCD initialise MPI will be beneficial at least for the design of some programs (CVC for example).

@sunpho84 How would this change affect Nissa? I would like to find a compromise which works for all (potential) client codes, but someone needs to call MPI_Init. There's a change about to land which allows runtime selection of the MPI thread level.

### DIFF
--- a/include/tmLQCD.h
+++ b/include/tmLQCD.h
@@ -41,6 +41,11 @@ extern "C"
   typedef struct {
     unsigned int nproc, nproc_t, nproc_x, nproc_y, nproc_z, cart_id, proc_id, time_rank, omp_num_threads;
     unsigned int proc_coords[4];
+#ifdef TM_USE_MPI
+    MPI_Comm cart_grid;
+#else
+    int cart_grid;
+#endif
   } tmLQCD_mpi_params;
 
   int tmLQCD_invert_init(int argc, char *argv[], const int verbose, const int external_id);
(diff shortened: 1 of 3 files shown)


kostrzewa:
We can also declare that the client code takes responsibility of initialising MPI. Ideally we would then have the MPI thread level passed down through the interface.

sunpho84:
According to the MPI standard, MPI_Init should be called as soon as possible, as you note in bff0e61. This means that the client is forced to call tmLQCD_invert_init before doing anything else.

I could certainly rearrange the initialization of the nissa library to cope with this, but what would happen if every linked library claimed the same prerogative...? 

Why is initializing MPI inside tmLQCD needed?

kostrzewa:
> Why is initializing MPI inside tmLQCD needed?

It's not and I would prefer not to do it. In fact, I'm going to revert the change. It was a misunderstanding in the CVC source code.

There might be some subtlety with the new thread level settings, but I will solve that differently.

kostrzewa:
Ugh, this is problematic after all. We need to initialise QMP for QPhiX to work. Perhaps I should expose the initialisation function and the client code should then call this instead of MPI_Init.

sunpho84:
Can you summarize in which order are things initialized currently, and which are the unavoidable dependency?

kostrzewa:
In order to use QPhiX, MPI initialisation goes via QMP. In order to avoid this, the client code would have to initialise QMP...

kostrzewa:
To answer your question in more detail:

init_parallel_and_read_input

initialises QMP or MPI (wihch one is used is chosen via a preprocessor macro defined in the installed qphix_config.h). Immediately after that the input file is read and init_omp() is called to set the number of threads. The ordering of the latter is because the number of threads is given in the input file and I wanted to get everything to do with parallelisation into a single routine. It can be split back up, however.

What is unavoidable is the choice between MPI and QMP.

PR #446, opened 2018-11-10
some generalisations for lib_wrapper

kostrzewa:
This will allow for much more straightforward use of the inverter interface by giving users the possibility to set up one or a few dummy operators (of the right class and with the right solver parameters) in the input file and then giving client code the ability to modify these at runtime.


kostrzewa:
This now also includes a ```tmLQCD_invert_eo``` routine which allows the solver for a particular operator to be called with the propagator and source specified in odd / even format.

kostrzewa:
@marcus-petschlies I guess this stuff will also be interesting for CVC.

kostrzewa:
@sunpho84 
okay, with ```tmLQCD_invert_qphix_direct``` and ```tmLQCD_[get,set]_op_params``` it should now be possible to do the inversions using the QPhiX double-single or double solver as an inner solver for your quadruple precision solver

kostrzewa:
Note, however, that we work in kappa normalisation such that a rescaling will be necessary before entering and after leaving the inner solver...

sunpho84:
ok, we can multiply or divide by 2k the source or solution. Wha t is this `tmLQCD_invert_qphix_direct` supposed to do? does it bypass the usual control through the operators?

kostrzewa:
@sunpho84 after working out the input file with Marco today I checked what we can really gain by using QPhiX as an inner solver. While in the regime of O(1000) iterations it's about a factor of 4 faster than Nissa, the gain in the charm+ region is not worth the effort, I think. That is, unless the situation on the 64c128 lattice is different. The speedup drops to a measly factor of 2 because we have to accomodate DDalphaAMG and so can't run with the most efficient setup of 8 MPI tasks per node and 8 threads per task. Also, as reported previously, at 27 nodes we are already hit by the scaling problems on KNL, which is a shame. I know that the kernels themselves are capable of at least a factor of four more... I still hope that huge pages and threaded comms may help at some point...

kostrzewa:
This will be merged via the https://github.com/etmc/tmLQCD/tree/get_cvc_to_work branch

PR #450, opened 2018-11-25
QUDA interface / lib_wrapper work

kostrzewa:
Making progress on the interface with the things that we need in order to run the various cpff tasks for the flavor_singlet project.


kostrzewa:
@marcus-petschlies This is the branch to target for CVC. Let's please coordinate much more closely how we progress.

PR #460, opened 2019-04-16
work on lib_wrapper / QUDA interface and CVC integration

kostrzewa:


kostrzewa:
It seems that I've set up the parameter input for coarse grid deflation correctly, but there are still a number of things that need to be upgraded in the input file parser to have more fine-grained control over several per-level parameters. I'm also not quite sure yet whether it is actually doing anything or if I still have to mess around with some things. Help with testing/exploration would be welcome!

kostrzewa:
I got the coarse-grid deflation working. Poly acceleration doesn't really work yet on my test ensemble, but I will try on the physical point lattice to see if I can make it purr there based on the settings by Kate. One thing to note: the input complexity for this thing is now way beyond reasonable...

kostrzewa:
This is now synced with the master branch. The effect of the modifications on the HMC can thus be tested.

io/gauge_read.c
```diff
@@ -185,10 +185,14 @@ int read_gauge_field(char * filename, su3 ** const gf) {
 
   g_update_gauge_copy = 1;
 
+  if(gf==g_gauge_field){
+    // reading a new gauge configuration moves the gauge_id a long way
+    // to guarantee that the change is propagated 
+    update_tm_gauge_id(&g_gauge_state, TM_GAUGE_PROPAGATE_THRESHOLD);
 #ifdef DDalphaAMG
-  if(gf==g_gauge_field)
```
kostrzewa, line 189: this definitely should NOT be removed...

kostrzewa: can someone else please check this point?

PR #469, opened 2019-06-14
bug fix: wrong plaquette in metadata just after reversibility check

palao:
The xlf-info LIME record contained a wrong value of the plaquette when the reversibility check occurs.
To reproduce it, take for instance "sample-input/sample-hmc0.input" and set (to make it simpler and faster) for instance:
Measurements = 11
ReversibilityCheckIntervall = 10

The plaquette value in the "xlf-info" header of the resulting conf is wrong (it corresponds to the plaquette computed on the reversed trajectory).
I checked that the averaged plaquette computed and the value written in "output.data" agree and both disagree with the value in the "xlf-info".

### DIFF
--- a/update_tm.c
+++ b/update_tm.c
@@ -188,7 +188,7 @@ int update_tm(double *plaquette_energy, double *rectangle_energy,
     }
     if(accept) {
       /* save gauge file to disk before performing reversibility check */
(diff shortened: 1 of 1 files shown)


kostrzewa:
Agreed, this was not correct!

kostrzewa:
thanks David!

PR #473, opened 2019-11-11
useful backports from quda_work branch

kostrzewa:
mainly, this changes `config.h` -> `tmlqcd_config.h` and installs this file in `include` in the build directory so it can be easily externally included as it should be

secondly, for some reason we lost 

```
<TMSOLVER,CSWSOLVER>{
  mg {
    optr->solver=MG;
    if(myverbose) printf("  Solver set to MG line %d operator %d\n", line_of_file, current_operator);
    BEGIN(name_caller);
  }
}
```

from `read_input.l`. This reintroduces it.


qphix_interface.cpp
```diff
@@ -73,7 +73,7 @@ extern "C" {
 #include <qphix/ndtm_reuse_operator.h>
 #include <qphix/ndtm_reuse_operator_clover.h>
 #include <qphix/print_utils.h>
-#include <qphix/qphix_config.h>
+#include <qphix/qphix_tmlqcd_config.h>
```
urbach: intended?

read_input.l
```diff
@@ -1331,6 +1331,14 @@ static inline double fltlist_next_token(int * const list_end){
...
 }
 
+<TMSOLVER,CSWSOLVER>{
+  mg {
+    optr->solver=MG;
+    if(myverbose) printf("  Solver set to MG line %d operator %d\n", line_of_file, current_operator);
+    BEGIN(name_caller);
+  }
+}
+
```
urbach: how could we loose this?

wrapper/lib_wrapper.c
```diff
@@ -201,7 +202,10 @@ int tmLQCD_invert_init(int argc, char *argv[], const int _verbose, const int ext
 #  endif
   }
 #endif
-  tmLQCD_invert_initialised = 1;  
+  tmLQCD_invert_initialised = 1;
+#ifdef TM_USE_MPI
+  MPI_Barrier(MPI_COMM_WORLD);
+#endif
```
urbach: why is this needed?

read_input.l
kostrzewa: I honestly have no idea...

qphix_interface.cpp
kostrzewa: nope, thanks for the catch!

wrapper/lib_wrapper.c
kostrzewa: Strictly speaking it is not, but when using MPI, we expose `g_cart_grid` via `tmLQCD_get_mpi_params` (such that external codes can `MPI_Comm_dup` this). I would like to make sure that we don't have any nasty surprises due to lack of synchronisation. It doesn't cost anything at this stage so I though why not.

kostrzewa:
Thanks for the review, always helpful to have more eyes on code... I've fixed the erroneous replacements.

urbach:
> Thanks for the review, always helpful to have more eyes on code... I've fixed the erroneous replacements.

thanks, good to be merged!

PR #477, opened 2020-08-04
fix compilation on Hawk

kostrzewa:
Using gcc 9.2.0 revealed some issues with circular dependencies in header files as well as problems dealing with manipulations that https://github.com/usqcd-software/c-lime does to preprocessor macros coming from its own `AC_INIT` and which are included via `lime.h`. (basically things like `PACKAGE_VERSION` are undef'ed).

I've cleaned up many of the include satements in the code and have attempted a workaround for the problem with c-lime.


tmlqcd_config.h
```diff
@@ -0,0 +1,29 @@
...
+
+ // include the autoconf-generated header
+ #include "tmlqcd_config_internal.h"
+ static const char* const TMLQCD_PACKAGE_BUGREPORT = PACKAGE_BUGREPORT;
+ static const char* const TMLQCD_PACKAGE_NAME = PACKAGE_NAME;
+ static const char* const TMLQCD_PACKAGE_STRING = PACKAGE_STRING;
+ static const char* const TMLQCD_PACKAGE_TARNAME = PACKAGE_TARNAME;
+ static const char* const TMLQCD_PACKAGE_VERSION = PACKAGE_VERSION;
+
+ // unlike lime, we don't undef anything
```
urbach, line 27: but you undef above...?

urbach (approved):
why does #include <tmlqcd_config.h> not work anymore?

tmlqcd_config.h
kostrzewa: I follow the convention that lime uses too.

kostrzewa: our definitions are in `tmlqcd_config_internal.h` which, importantly, does not have header guards

kostrzewa: or rather, it *shouldn't* have header guards, unfortunately I seem to have messed that up :)

kostrzewa:
The problem is one of include order and what other autoconf-based packages do with defines like `PACKAGE_VERSION`. C-lime, for example, undefs these and then they are gone.

Now, one would think that one could simply include `tmlqcd_config.h` first and then `lime.h`, but unfortunately, that doesn't always help because `tmlqcd_config.h` might have been included somewhere, where `lime.h` was NOT included. As a result, `_CONFIG_H` is now defined. When the compiler then gets to a file where`lime.h` is actually included, the `PACKAGE_VERSION` definition will be undeffed by `lime.h` and won't be available anymore.

Hence the deferral to an include file without header guards and the redefinition.

tmlqcd_config.h
kostrzewa: Hence, by storing our `PACKAGE_` strings in separate variables, we can be sure that no other package overrides (or undefs) them.

PR #479, opened 2021-02-22
Gamma doc

kostrzewa:
I think this should be correct now


kostrzewa:
Note that this finally makes the code and the documentation consistent again (sorry about that)...

image

image

urbach (approved):

PR #483, opened 2021-04-01
Renaming structures and added extern, removed Direction to avoid compile clash

### LINKED ISSUE #481: Multiple inclusions of block.h
When linking I get the following error
```

/usr/bin/ld: ./lib/libhmc.a(invert_clover_eo.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../block.h:106: multiple definition of `blk_gauge_eo'; ./lib/libhmc.a(operator.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../block.h:106: first defined here

/usr/bin/ld: ./lib/libhmc.a(invert_clover_eo.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../block.h:105: multiple definition of `nblks_dir'; ./lib/libhmc.a(operator.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../block.h:105: first defined here

[...]

```

and so on. It looks like `block.h` contains a number of global variables, some of them are referred as `extern`, some not, as it is the case for `blk_gauge_eo` and the others. The `block.h` header is included inside `clovertm_operators.h`, which in turns is included in many place. This gives rise to the clash.

Is this somehow new issue? I am in the quda_work branch, but it looks to me this is generic.

sunpho84:
fix #481 and  #485 #484


sunpho84:
I've included the commits needed to fix #484

Issue #485, opened 2021-04-01
Multiple definitions of structure Direction in little_field_gather.c and little_D.c


sunpho84:
```
/usr/bin/ld: ./lib/libxchange.a(little_field_gather.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/xchange/../../xchange/little_field_gather.c:47: multiple definition of `Direction'; ./lib/libhmc.a(little_D.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../little_D.c:71: first defined here
```

a `Direction` structure is indeed defined in both `little_field_gather.c` and `little_D.c`, but in the second is never referred, so I guess it can be removed

urbach:
are you sure that this is the latest version of the DD solver? I have the feeling that I fixed these problems already at some point?

> 
> ```
> /usr/bin/ld: ./lib/libxchange.a(little_field_gather.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/xchange/../../xchange/little_field_gather.c:47: multiple definition of `Direction'; ./lib/libhmc.a(little_D.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../little_D.c:71: first defined here
> ```
> 
> a `Direction` structure is indeed defined in both `little_field_gather.c` and `little_D.c`, but in the second is never referred, so I guess it can be removed
> 
> -- 
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly or view it on GitHub:
> https://github.com/etmc/tmLQCD/issues/485

-- 
Carsten Urbach
e-mail: ***@***.***
        ***@***.***
Fon   : +49 (0)228 73 2379
skype : carsten.urbach
URL: http://www.carsten-urbach.eu

kostrzewa:
This may very well be. https://github.com/etmc/tmLQCD/pull/460 has been waiting for a review since April 2019 and there are conflicts in various files indicated in that pull.

kostrzewa:
I think we need to merge master into quda_work first and then see from there... I'm afraid the number of conflicts is really high.

kostrzewa:
@urbach the multiple definitions stuff definitely is still a problem in the master branch, it's just "by accident" that the issue was not triggered by compilers prior to GCC 10 apparently...

PR #487, opened 2021-04-06
Additional Direction not needed

sunpho84:
Solving this additional error, similar to #485 

```
/usr/bin/ld: ./lib/libxchange.a(little_field_gather.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/xchange/../../xchange/little_field_gather.c:47: multiple definition of `Direction'; ./lib/libhmc.a(block.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../block.c:78: first defined here
```

### DIFF
--- a/block.c
+++ b/block.c
@@ -65,17 +65,6 @@ _Complex float * little_A_eo_32 = NULL;
 int * block_idx;
 int * block_evenidx;
 int * block_oddidx;
-enum{
-  NONE = 0,
-  T_UP = 1,
-  T_DN = 2,
-  X_UP = 3,
-  X_DN = 4,
-  Y_UP = 5,
-  Y_DN = 6,
-  Z_UP = 7,
-  Z_DN = 8
-} Direction;
 
 static void (*boundary_D[8])(spinor * const r, spinor * const s, su3 *u) =
 {boundary_D_0, boundary_D_1, boundary_D_2, boundary_D_3, boundary_D_4, boundary_D_5, boundary_D_6, boundary_D_7};


kostrzewa:
In the current `quda_work` branch, there's only a single definition of this enum in block.c. Have you perhaps changed things locally?

kostrzewa:
Ah, no, I see, there's another one in `little_field_gather.c`, but with a different numbering!

kostrzewa:
`block.c`:

```
enum{
  NONE = 0, 
  T_UP = 1, 
  T_DN = 2, 
  X_UP = 3, 
  X_DN = 4, 
  Y_UP = 5, 
  Y_DN = 6, 
  Z_UP = 7, 
  Z_DN = 8
} Direction;                                                                                                                               
```
`xchange/little_field_gather.c`:

```
enum{
  T_UP = 0,
  T_DN = 1,
  X_UP = 2,
  X_DN = 3,
  Y_UP = 4,
  Y_DN = 5,
  Z_UP = 6,
  Z_DN = 7 
} Direction;                                                                                                                               
```

kostrzewa:
`NONE` is never used apparently, but I honestly can't tell if this won't break things. It's not like we need a Lüscher-style deflated solver given the availability of DDalphaAMG. I thus don't much care if this breaks, but @urbach should decide since he spent so much time fixing this stuff...

sunpho84:
It looks to me like the `Direction` structure in `block.c` is simply dead code, as it is not referred anywhere in the `TU`, so I would be tempted to say that removing it should have no impact. 

Indeed both are contained in a `.c` file, and `block.c` compiles no matter the presence of the structure. Am I too naive...?

kostrzewa:
I will just merge this in now. If it breaks things, somebody else will have to fix it...

sunpho84:
I'll keep my fingers crossed...

PR #488, opened 2021-04-06
no_eigenvalues already defined extern in the related header

sunpho84:
Fix this
```
/usr/bin/ld: ./lib/libsolver.a(eigenvalues.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/solver/../../solver/eigenvalues.c:61: multiple definition of `no_eigenvalues'; ./lib/libhmc.a(read_input.o):/home/francesco/QCD/SORGENTI/tmLQCD/build_quda/../read_input.l:222: first defined here
```

### DIFF
--- a/solver/eigenvalues.c
+++ b/solver/eigenvalues.c
@@ -58,7 +58,7 @@ double * eigenvls = NULL;
 double max_eigenvalue;
 double * inv_eigenvls = NULL;
 int eigenvalues_for_cg_computed = 0;
-int no_eigenvalues, evlength;
+int evlength;
 
 /* the folowing two are needed for the overlap */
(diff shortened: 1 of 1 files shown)


kostrzewa:
Argh, this is super nasty. Variables with this name are used (and modified) in various places throughout the codebase and I don't think that the various instances of this variable are supposed to mean the same thing...

kostrzewa:
It's also used as global input AND output ... soooo nasty.

sunpho84:
Very good... so my naive fix is simply avoiding the error but not changing the behaviour, as I guess the compiler was somehow melding the two instances as if one was defined with extern linkage. 

Is it somehow clear how to proceed?

PR #491, opened 2021-07-23
QUDA inverter in the HMC

Marcogarofalo:
tentative to use quda inverter in hmc.


quda_interface.c
```diff
@@ -1416,3 +1416,94 @@ void _setQudaMultigridParam(QudaMultigridParam* mg_param) {
...
+                            max_iter);
+  // overwriting  inv_param set by _setOneFlavourSolverParam
+
+  //solution_type    solve_type    Effect
+  //  -------------    ----------    ------
+  //  MAT              DIRECT        Solve Ax=b
+  //  MATDAG_MAT       DIRECT        Solve A^dag y = b, followed by Ax=y
+  //  MAT              NORMOP        Solve (A^dag A) x = (A^dag b)
+  //  MATDAG_MAT       NORMOP        Solve (A^dag A) x = b
+  inv_param.solution_type = QUDA_MATDAG_MAT_SOLUTION ;
```
kostrzewa, line 1475: I think we need `QUDA_MATPCDAG_MATPC_SOLUTION` here

Marcogarofalo: with that, I get an error in the run `Source has zero norm`
```
# TM_QUDA: mu = 0.036231884058, kappa = 0.138000000000, csw = 1.000000000000
# TM_QUDA: time spent in reorder_spinor_toQuda: 0.000079 secs
# QUDA: ERROR: Source has zero norm (rank 0, host lnode08, /qbigwork2/bartek/code/quda_1.1.x_8efd5cfade419574bf4710545a07c08824b2c1c2/lib/interface_quda.cpp:2942 in invertQuda())
# QUDA:        last kernel called was (name=N4quda4blas5Norm2IddEE,volume=2x4x4x4,aux=vol=128,stride=128,precision=8,order=2,Ns=4,Nc=3,TwistFlavour=1,nParity=1)
```
maybe it is due to the Odd part of the source which is initialised with
spinor** tempE;
 init_solver_field(&tempE, VOLUMEPLUSRAND/2, 1);

Marcogarofalo: if I invert even and odd in the source 
```
iteration_count= invert_eo_MMd_quda(P,tempE[0],  temp[0], tempE[0], ...
```
I get a different error 
```
# QUDA: ERROR: Preconditioned solution requires a preconditioned solve_type
```

so there is a reordering problem plus something else

kostrzewa: `QUDA_NORMOP_PC_SOLVE` would be the right `solve_type` for this situation

kostrzewa: I think it should be relatively straightforward to make this work in the end. QUDA expects all the sites from the checkerboard that we want to solve on (either even or odd, in our case odd) at the beginning of the spinor that is passed.

I think I see two options

1) add another parameter to `reorder_spinor_fromQuda` and `reorder_spinor_toQuda`, "`int cb`", for example, which specifies which part of the spinor is non-zero and should be put at the beginning of the output or which sits at the beginning of the input.
2) add two more functions, `reorder_eo_spinor_toQuda` and `reorder_eo_spinor_fromQuda` which perform the appropriate reordering for a single checkerboard only (reducing the overhead in the process)
  i) This would then allow the `invert_eo_MMd_quda` function to have an interface with a single input and a single output.

I think I would prefer option (2) as the resulting code duplication is quite minimal.

What do you think?

Marcogarofalo: - I implement reorder_eo_spinor_toQuda and fromQuda. 
- I remove convert_eo_to_lexic since now everything is done in reorder_eo_spinor_toQuda. 
- the result is the same as the one I was getting with the previous  Even and Odd input  `invert_eo_MMd_quda`. However, it is still different from the host only reference 
```
diff onlinemeas.000000 ../run_reference/onlinemeas.000000
1,9c1,9
< 1  1  0  3.381291e+01  0.000000e+00
< 1  1  1  2.191772e+00  1.940924e+00
< 1  1  2  4.228271e-01  0.000000e+00
< 2  1  0  2.090087e+00  0.000000e+00
< 2  1  1  1.466976e+00  -1.294756e+00
< 2  1  2  3.815794e-04  0.000000e+00
< 6  1  0  9.620622e-01  0.000000e+00
< 6  1  1  8.114292e-02  -7.207055e-02
< 6  1  2  -1.133856e-03  0.000000e+00
---
> 1  1  0  3.525860e+01  0.000000e+00
> 1  1  1  2.364022e+00  2.464564e+00
> 1  1  2  5.648465e-01  0.000000e+00
> 2  1  0  -7.551440e-01  0.000000e+00
> 2  1  1  1.517322e+00  -1.495484e+00
> 2  1  2  8.362970e-03  0.000000e+00
> 6  1  0  4.999061e-01  0.000000e+00
> 6  1  1  4.593042e-02  -1.548241e-01
> 6  1  2  -5.299438e-02  0.000000e+00

```

kostrzewa: Are the results still the same with the updated `reorder_spinor_eo_toQuda` and its inverse?

The differences that you see in the PP, PA and AA correlators are not very large, however. On this kind of small lattice it's entirely possible to very quickly have large differences to a reference run using different solvers in the evolution.

In any case, the config was likely not accepted (since the residuals, at least for me, are huge).

kostrzewa:
@Marcogarofalo @sunpho84 this now has the first working call from `solve_degenerate` to the QUDA solvers. I think based on this it should now be mostly a matter of setting parameters properly to also get CLOVERDET running without mass preconditioning (there's an issue with reloading the clover field wihch I never resolved...).

For the determinant ratios
1) The correct usage of the `mu2` and `kappa2` parameters for DETRATIO must be ensured
2) Work in QUDA is required to support CLOVERDETRATIO due to its usage of the `rho` parameter. This is also required to use mass preconditioning in CLOVERDET.

For the RHMC, an interface to QUDA's multi-shift solver is required and the rescaling that is also done in the QPhiX interface.

Because the reordering routines will have some overhead, support should be added to QUDA to perform the gamma basis change on the device as well as the ability to load and store tmLQCD fields directly. See https://github.com/qcdcode/quda/issues/13

kostrzewa:
To use the MG in the HMC, a lot more work is required as one needs to do setup evolution. There's a test in QUDA's test directories which demonstrates how this is done.

kostrzewa:
Just as a note: the MG solver "works" in this way, but I'm almost certain that it's not correct to use it like this (in terms of getting a good algorithm). I think the MG-preconditioner actually isn't doing anything to precondition the fine system. The number of outer iterations is currently at the level of a few hundred on a 24c32 lattice with `2kappamu = 0.1`, which is simply a very poor GCR ;)

Marcogarofalo:
> @Marcogarofalo @sunpho84 this now has the first working call from `solve_degenerate` to the QUDA solvers. I think based on this it should now be mostly a matter of setting parameters properly to also get CLOVERDET running without mass preconditioning (there's an issue with reloading the clover field wihch I never resolved...).

I think that the CLOVERDET monomial works fine in quda. I generate two configurations with sample-hmc-tmcloverdet.input with the modification: 
```
BeginMonomial CLOVERDET
  Timescale = 1
  2KappaMu = 0.01
  rho = 0.0
  CSW = 1.00
  kappa = 0.138
  AcceptancePrecision =  1.e-20
  ForcePrecision = 1.e-12
  Name = cloverdet
  useexternalinverter = quda
  usesloppyprecision = single
#  usecompression = 12
#  solver = CG
EndMonomial
```
the gauge configurations are the same up to 1e-6 and the only difference in the online measuraments is 

```
diff onlinemeas.000002 ../run_reference/onlinemeas.000002_tmcloverdet
8c8
< 6  1  1  9.685959e-02  -3.568352e-02
---
> 6  1  1  9.685960e-02  -3.568352e-02
```
 
> For the RHMC, an interface to QUDA's multi-shift solver is required and the rescaling that is also done in the QPhiX interface.
> 
> Because the reordering routines will have some overhead, support should be added to QUDA to perform the gamma basis change on the device as well as the ability to load and store tmLQCD fields directly. See [qcdcode/quda#13](https://github.com/qcdcode/quda/issues/13)

adding the tmLQCD gamma matrices in quda is not an option?

kostrzewa:
> I think that the CLOVERDET monomial works fine in quda. I generate two configurations with sample-hmc-tmcloverdet.input with the modification:

Yes, it works. The problem is that for the next trajectory, reloading the clover field hits an issue with the parameter struct which I know about but never resolved because we usually don't do multiple configs per run in analysis.

kostrzewa:
> adding the tmLQCD gamma matrices in quda is not an option?

many parts in QUDA are hard-coded for algorithmic reasons: the DeGrand-Rossi basis is used in the MG (it's very similar to ours) because it's chiral (and this allows the chiralities to be split on the coarse grid) and the UKQCD (non-relativistic) basis is used elsewhere

the trick is to add the correct reorderings to QUDA (all the machinery is there). On the GPU, the reordering before a solve, for example, is a tiny overhead compared to actually sending the field to the device.

kostrzewa:
please note https://github.com/etmc/tmLQCD/issues/494

urbach:
Is there anything speaking against merging this in and work directly on `tmLQCD:quda_work_add_actions`?

kostrzewa:
> Is there anything speaking against merging this in and work directly on tmLQCD:quda_work_add_actions?

I wanted to give everyone a chance to follow what's happening by looking at the updates in this PR. I would say at this point we should merge it in and continue in a new PR (not in Marco's fork but directly in etmc/tmLQCD)

PR #492, opened 2021-07-26
sudo apt-get update && sudo apt-get install ...

Marcogarofalo:
it should fix the error in the action workflow


kostrzewa:
thanks!

PR #498, opened 2021-08-10
Dash in filename

### LINKED ISSUE #478: parser cannot deal with minus signs in FILENAME
`GaugeConfigInputFile = this-is-a-dir/conf` does not work

urbach:
should fix #478


kostrzewa (approved):
works as expected

PR #499, opened 2021-08-11
WIP PR to continue the work of getting the HMC to run fully on GPUs

kostrzewa:


Marcogarofalo (approved):

kostrzewa:
Alright, I'll merge this in and then the next big quality-control step that I will do is to run some CVC jobs (using my cvc_depgraph fork) which rely on tmLQCD to interface with QUDA, just to test that we didn't break any expected behaviour.

PR #502, opened 2021-08-30
Split gauge update

urbach:
some steps towards gauge update on GPUs


misc_types.h
```diff
@@ -89,6 +89,13 @@ typedef enum ExternalInverter_s {
   QPHIX_INVERTER
 } ExternalInverter;
 
+/* enumeration type for the external inverter */
```
kostrzewa, line 92: @urbach does it perhaps make sense to merge this and `ExternalInverter` ?

urbach:
> @kostrzewa commented on this pull request.
> 
> 
> 
> > @@ -89,6 +89,13 @@ typedef enum ExternalInverter_s {
>    QPHIX_INVERTER
>  } ExternalInverter;
>  
> +/* enumeration type for the external inverter */
> 
> @urbach does it perhaps make sense to merge this and `ExternalInverter` ?

thought about it, but then decided against it.

ExternalLibrary is going to put the whole GaugeUpdate on the GPU, I
hope. In the other case only the inverter is on the GPU. So, in my
opinion two very different things. However, I can easily change this,
if there are good arguments.

misc_types.h
kostrzewa: > thought about it, but then decided against it.

> ExternalLibrary is going to put the whole GaugeUpdate on the GPU, I hope. In the other case only the inverter is on the GPU. So, in my opinion two very different things. However, I can easily change this, if there are good arguments.

Agreed.

kostrzewa: the comment above should be adapted

kostrzewa: It also might be that we will need the generality to use two different external libraries at the same time: one for the solver and one for the remaining tasks in the deriv and/or hb and/or acc steps.

monomial/monomial.h
```diff
@@ -111,7 +111,7 @@ typedef struct {
   double PrecisionHfinal;
   double StildeMin, StildeMax;
   double EVMin, EVMax, EVMaxInv;
-  ExternalLibrary external_library;
+  ExternalLibrary use_external_lib;
```
kostrzewa, line 114: to be consistent with `external_inverter` in `solver_params_t`, could we use only `external_lib` here as a variable name? This is also consistent with the variable name for the default.

kostrzewa:
@urbach should we split the work on this somehow? In principle I have some time available in the next two weeks here or there.

urbach:
> @urbach should we split the work on this somehow? In principle I have some time available in the next two weeks here or there.

sure that would be great. I need to understand first, what the QUDA
routine actually does. Maybe @sbacchio can help there?

sbacchio:
Hi yes I can also help. Do you want to have any discussion on how to distribute it? and what you have in mind for now?

urbach:
> Hi yes I can also help. Do you want to have any discussion on how to distribute it? and what you have in mind for now?

discussion would be good. Bartek has certainly more inside already, for the following steps are needed/unclear

- what exactly does the QUDA function `computeGaugeForceQuda` and how
  to call it
- as far as I understand `computeGaugeForceQuda` updates the momenta
  directly. How to join this with the tmLQCD momenta?
- do we want to start with forking off in `gauge_derivative` first and
  then generalise when it is working, or do we start directly with a
  more general interface function?

urbach:
what about a short meeting today at 1pm or 2pm?

sbacchio:
Both timings would be fine for me

kostrzewa:
Same here although I would slightly prefer 2pm

urbach:
> Same here although I would slightly prefer 2pm

then let's say 2pm.

sunpho84:
Hi

sorry I'm late, where can I join?

kostrzewa:
Just doing some dummy work to see if I can easily get a working call to the QUDA gauge force set up. Refinements can be made later (such as doing the projection onto the adjoint rep already in QUDA).

sbacchio:
Hi Bartek, just for not doing the work twice, we are working on it... We have almost everything figured out and at some point early next week we are going to push here the implementation. @pittlerf is working on it with me

kostrzewa:
Alright, good to know. I'll push my latest additions (which still result in a segfault unfortunately) in a separate branch and you can then force-push here if you'd like.

sbacchio:
ok great!

kostrzewa:
Would be good to have some WIP commits to look at if possible.

sbacchio:
ok we have clean up to do, later today..

kostrzewa:
no worries, I didn't mean to imply any pressure, I'm just really interested in what you came up with

sbacchio:
@kostrzewa can we be added to etmc so we can push here directly?
At the moment changes are under https://github.com/pittlerf/tmLQCD/tree/testing_gauge_force_ferenc

kostrzewa:
@pittlerf was already a member and I've just added you, @sbacchio

kostrzewa:
Thanks @pittlerf and @sbacchio. Needs some minor conflict resolution

quda_interface.c
```diff
@@ -2210,3 +2317,232 @@ int invert_eo_quda_twoflavour_mshift(spinor ** const out_up, spinor ** const out
...
+    // Plaq coeffs
+    if(i<6)
+      loop_coeff[i] = -1.5 * g_beta * mnl->c0;
+    // Rect coeffs
+    else
+      loop_coeff[i] = -1.5 * g_beta * mnl->c1;
+  }
+  
+  // prepares gauge_quda
+  reorder_gauge_toQuda(hf->gaugefield, 18);
```
kostrzewa, line 2529: While I agree with this to be general, this modifies the state of `gauge_quda`, ignoring the state tracking. As a result, `g_gauge_state.gauge_id` and `quda_gauge_state.gauge_id` will not reflect the true state of the two gauge fields. Currently we don't have any situation in which `hf->gaugefield` does not correspond to `g_gauge_field` (although in the future one might of course have, for example, a smeared gauge field for fermionic monomials). For now I would feel better if `_loadGaugeQuda` could be generalized to take an argument or if `g_gauge_field` would simply be assumed. I don't like either option very much but I think it's better than losing track of the state.

sbacchio:
We checked serially the current implementation and as far as we can tell it matches the gauge_derivative implementation. 
Now we have some issue running in parallel.. It dies in the initialization of QUDA.. still have to investigate further

quda_interface.c
kostrzewa: I'm also in favour of using the names of enum values wherever possible rather than hard-coding integers.

kostrzewa:
> We checked serially the current implementation and as far as we can tell it matches the gauge_derivative implementation. Now we have some issue running in parallel.. It dies in the initialization of QUDA.. still have to investigate further

It (e.g. tmLQCD's `hmc_tm` or `invert` or `offline_measurement`) launches fine in parallel for me, but I have some `dH` issues when the gauge monomial is on the GPU (which are likely to do with the fact that I forgot to pull in the changes to QUDA that you've made w.r.t. to the `double` alignment). The lib is recompiling as we speak and I'll test again in a few minutes.

quda_interface.c
```diff
@@ -2210,3 +2317,232 @@ int invert_eo_quda_twoflavour_mshift(spinor ** const out_up, spinor ** const out
...
+    }
+  
+#ifdef TM_USE_OMP
+  }
+#endif
+  tm_stopwatch_pop(&g_timers, 0, 0, "TM_QUDA", __func__);
+}
+
+
+void compute_gauge_force_quda(monomial * const mnl, hamiltonian_field_t * const hf) {
```
kostrzewa, line 2493: Naming this is my fault, of course ;) I wanted to suggest that we rename this to `compute_gauge_derivative_quda` just to be consistent with the tmLQCD nomenclature.

quda_interface.c
```diff
@@ -2210,3 +2317,232 @@ int invert_eo_quda_twoflavour_mshift(spinor ** const out_up, spinor ** const out
...
+      { 7, 4, 0 },
+      { 1, 4, 6 },
+      { 6, 4, 1 },
+      { 2, 4, 5 },
+      { 5, 4, 2 } } 
+  };
+
+
+void add_mom_to_derivative(su3adj** der) {
+  // mom_quda -> mom_quda_reordered
```
kostrzewa, line 2459: this comment can be removed

kostrzewa:
I'm afraid that also with the additional commits in `feature/ndeg-twisted-clover` I see large `dH` in a pure-gauge run using the QUDA version of the gauge force.

sbacchio:
Thanks for checking, but yes it is still WIP.. so far we did unit tests with our executable.

pittlerf:
> We checked serially the current implementation and as far as we can tell it matches the gauge_derivative implementation.
> Now we have some issue running in parallel.. It dies in the initialization of QUDA.. still have to investigate further

Now this is solved, for we had problems running interactively, using script it works also parallel

pittlerf:
> I'm afraid that also with the additional commits in `feature/ndeg-twisted-clover` I see large `dH` in a pure-gauge run using the QUDA version of the gauge force.

Hi Bartek, could we have your input script, we have implemented some checks and would like to try.

kostrzewa:
Sure, I just do something like:

```
L=16
T=32

nrxprocs = 1
nryprocs = 1
nrzprocs = 1
ompnumthreads = 3

Measurements = 10000
StartCondition = hot

NSave = 500000
ThetaT = 1.0

UseEvenOdd = yes
ReversibilityCheck = no
ReversibilityCheckIntervall = 100
DebugLevel = 2

BeginMonomial GAUGE
  Type = Iwasaki
  beta = 1.9
  Timescale = 0
  UseExternalLibrary = quda
EndMonomial

BeginIntegrator 
  Type0 = 2MN
  IntegrationSteps0 = 100
  Tau = 1.0
  Lambda0 = 0.193
  NumberOfTimescales = 1
EndIntegrator
```

for pure gauge.

kostrzewa:
You should be able to push. How have you configured the remote?

kostrzewa:
Actually you are right, the base permission was "read" for some reason. You should be able to push now.

pittlerf:
> Actually you are right, the base permission was "read" for some reason. You should be able to push now.

thank you :)

quda_interface.c
```diff
@@ -2459,16 +2459,11 @@ void add_mom_to_derivative(su3adj** der) {
...
-#pragma omp parallel
-  {
-#endif
-
 #ifdef TM_USE_OMP
 #pragma omp parallel for collapse(2)
 #endif
-  for( int mu=0; mu<4; mu++ )
-    for( size_t v=0; v<VOLUME; v++ ){
+  for( size_t v=0; v<VOLUME; v++ )
```
kostrzewa, line 2465: I was wondering about this but it would not have led to correctness issues, just to slowness.

kostrzewa:
Very nice, the last commit has done the trick. I have a few more instrumentation which I'd like to push if you don't mind and one small performance improvement.

quda_interface.c
pittlerf: yeah, I think the removing of the outside OMP parallel somehow created the problem, the test_gauge_force exec seemed to be worked properly and we did not understand why hmc_tm lead to different results, then I have checked that in hmc_tm I used 3 OMP threads, and setting to 1 lead to the correct results

pittlerf:
> Very nice, the last commit has done the trick. I have a few more instrumentation which I'd like to push if you don't mind and one small performance improvement.

of coarse, go ahead :)

quda_interface.c
```diff
@@ -2520,16 +2520,18 @@ void compute_gauge_force_quda(monomial * const mnl, hamiltonian_field_t * const
   
   // prepares gauge_quda
   reorder_gauge_toQuda(hf->gaugefield, 18);
-
+  
+  #pragma omp parallel for
```
kostrzewa, line 2524: not sure if this actually helps but I expect that it does

kostrzewa:
Thanks! It would be great if we could resolve the merge conflict and go over the remaining issues over the next few days.

kostrzewa:
I think I've addressed the points in 6f579c8 and 7f61ee1.

kostrzewa:
I realised that the merge conflict was rather major so I went ahead and fixed it, it wasn't particularly clear...

pittlerf:
We added also an online checking tool, which can be used in debugging. Shall we remove the executable test_gauge_derivative then, or we keep it?

kostrzewa:
> We added also an online checking tool, which can be used in debugging. Shall we remove the executable test_gauge_derivative then, or we keep it?

Probably better to remove it since it might break and then we need to take care of it.

kostrzewa:
I've made some small adjustments and think that this can be merged. The gauge derivative regularly produces relative deviations in excess of `1e-10` on my machine, so I've increased the threshold. I don't think that the deviations that I see are worrisome, however as they occur only when the component in question is of a similar size to the threshold. I've made program termination dependent on `g_strict_residual_check`, which allows one to study the behaviour of the deviations on all lattice sites over mutliple trajectories.

PR #505, opened 2021-09-13
adding in which monomial the timing is measured

Marcogarofalo:
I introduce a global char * which is set to the working monomial in the update

### DIFF
--- a/Ptilde_nd.c
+++ b/Ptilde_nd.c
@@ -201,7 +201,7 @@ void Ptilde_ndpsi(spinor *R_s, spinor *R_c, double *dd, int n,
   free(auxc_);
   free(aux2c_);
   free(aux3c_);
-  tm_stopwatch_pop(&g_timers, 0, 1, "", __func__);
+  tm_stopwatch_pop(&g_timers, 0, 1, current_mnl, __func__);
 }
 
 double chebtilde_eval(int M, double *dd, double s){
@@ -350,6 +350,6 @@ void degree_of_Ptilde(int * _degree, double ** coefs,
   free(sc_);
   free(auxc_);
   free(aux2c_);
-  tm_stopwatch_pop(&g_timers, 0, 1, "", __func__);
+  tm_stopwatch_pop(&g_timers, 0, 1, current_mnl, __func__);
   return;
 }
--- a/Ptilde_nd.c
+++ b/Ptilde_nd.c
@@ -101,7 +101,7 @@ void Ptilde_cheb_coefs(double aa, double bb, double dd[], int n, double exponent
 
 void Ptilde_ndpsi(spinor *R_s, spinor *R_c, double *dd, int n, 
 		  spinor *S_s, spinor *S_c, matrix_mult_nd Qsq) {
-  tm_stopwatch_push(&g_timers);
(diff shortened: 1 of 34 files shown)


kostrzewa:
I would prefer *not* to have yet another global variable to take care of. I think it would be better to add arguments to `tm_stopwatch_push`, such that it can also print at some debug level (let's say `>= 3`).

Marcogarofalo:
Then how tm_stopwatch_push should know the monomial? The other option that I can think of is to pass the monomial name to all the functions like sw_spinor_eo, sw_all, deriv_Sb, ... 
Do you have something better?

kostrzewa:
It doesn't need to. The only thing that's required is that inside the monomial functions, one calls:

```
tm_stopwatch_push(&g_timers, 0, 3, mnl->name, __func__);
```

which then prints something like

```
# monomial_name: Entering __func__ level: %d
```

In the analysis of the output, the timings can then be associated to the correct context by reading `monomial_name` and the `level` and the final output of:

```
tm_stopwatch_pop(&g_timers, 0, 1, mnl->name, __func__);
```

Anything at levels higher than these two outputs belongs to that context.

kostrzewa:
Another option is to add an explicit "context" information for `tm_stopwatch_push`. In other words, the `g_timers` struct contains another array of strings (with some good length each, up to `500` or so).

```
tm_stopwatch_push(&g_timers, "context");
```

which then stores "context" in `g_timers.context[level]`. This context could be, for example, `mnl->name` and can even be used recursively in the logic (if the argument is empty, use the context of the previous level).

```
tm_stopwatch_push(&g_timers, "")
```

would thus result (internally) in

```
g_timers.context[level] = g_timers.context[level-1];
```

And then you can even have the context output at every call.

Marcogarofalo:
ok, then in a function like 
```
void deriv_Sb(const int ieo, spinor * const l, spinor * const k,
              hamiltonian_field_t * const hf, const double factor) 
```
I have to add 
```
void deriv_Sb(const int ieo, spinor * const l, spinor * const k,
              hamiltonian_field_t * const hf, const double factor, const char* mnl_name ) 
```

kostrzewa:
> I have to add

No, it's not necessary. When you have:

`Entering xxx in yyy, level: n`

in the output, anything *above* this level is automatically within that context. In other words: the `deriv_Sb` used within the monomial function will automatically have `level: n+1` and can be associated correctly at the analysis stage.

In addition, doing what I wrote above in https://github.com/etmc/tmLQCD/pull/505#issuecomment-918173628 would even give you the context locally.

kostrzewa:
To be more concrete, you could have:

```tm_stopwatch_push(&g_timers, mnl->name);``` 

at the top of `det_derivative`, for example. Internally, this sets

```timers->context[level] = "name of the nomial";```

Then, when `tm_stopwatch_push(&g_timers, "")` is called with an empty context (for example, inside `deriv_Sb`), this instead sets

```timers->context[level] = timers->context[level-1];```

which can then be printed as part of the `tm_stopwatch_pop` inside of the funtion is question. This way all `pop`s would be instrumented.

kostrzewa:
@Marcogarofalo can you push this branch to the etmc repo please?

PR #516, opened 2021-12-15
Quda work add actions wflow

Marcogarofalo:
Wflow is working but the output goes to stdout

### DIFF
--- a/meas/gradient_flow.c
+++ b/meas/gradient_flow.c
@@ -164,7 +164,7 @@ void gradient_flow_measurement(const int traj, const int id, const int ieo) {
 
   if (meas->external_library == QUDA_LIB) {
 #ifdef TM_USE_QUDA
-    compute_WFlow_quda( eps , tmax);
+    compute_WFlow_quda( eps , tmax, traj, outfile);
 #else
     fatal_error(
         "Attempted to use QUDA_LIB in gradient flow measurement but tmLQCD has been compiled "
--- a/meas/gradient_flow.c
+++ b/meas/gradient_flow.c
@@ -160,15 +160,11 @@ void gradient_flow_measurement(const int traj, const int id, const int ieo) {
     fprintf(outfile, "traj t P Eplaq Esym tsqEplaq tsqEsym Wsym Qsym\n");
   }
 
-  aligned_su3_field_t vt = aligned_su3_field_alloc(VOLUMEPLUSRAND+g_dbw2rand);
-  aligned_su3_field_t x1 = aligned_su3_field_alloc(VOLUMEPLUSRAND+g_dbw2rand);
-  aligned_su3_field_t x2 = aligned_su3_field_alloc(VOLUMEPLUSRAND+g_dbw2rand);
-  aligned_su3_field_t z = aligned_su3_field_alloc(VOLUME);
+  
 
   if (meas->external_library == QUDA_LIB) {
 #ifdef TM_USE_QUDA
     compute_WFlow_quda( eps , tmax);
-    // fatal_error("SIAMO ARRIVATI QUA!", __func__);
 #else
     fatal_error(
         "Attempted to use QUDA_LIB in gradient flow measurement but tmLQCD has been compiled "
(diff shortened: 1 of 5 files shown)


Marcogarofalo:
We do not know how to get in tmLQCD the variables like
`QudaGaugeObservableParam param` computed in the Wflow

Marcogarofalo:
This version works only with 
https://github.com/qcdcode/quda/commit/49a0ce12cafdc05092bc34c529f3cc30eb87f419

kostrzewa:
@Marcogarofalo thanks. Any changes to QUDA should be based on the develop branch though (which contains all of feature/ndeg-twisted-clover), just to make merging them easier.

Besides this, I think it would be better if the array of `QudaGaugeObservableParam` were created in tmLQCD (or any other external program using the lib) and passed to `performWFlownStep` (with the existing behaviour the default when one passes `nullptr`). In any case I doubt that Kate would accept the use of `new` and `delete`.

kostrzewa:
Ah, wait, I understand, sorry. Let's try it like this. I think, however, that `newQudaGaugeParam` should be called on all elements to make sure that any future changes don't break the initialisation.

Marcogarofalo:
> @Marcogarofalo thanks. Any changes to QUDA should be based on the develop branch though (which contains all of feature/ndeg-twisted-clover), just to make merging them easier.

ok I will try 

> 
> Besides this, I think it would be better if the array of `QudaGaugeObservableParam` were created in tmLQCD (or any other external program using the lib). In any case I doubt that Kate would accept the use of `new` and `delete`.

Since we could not implement a dofoult value for a parameter in c we add a function in quda performWFlownStep_param that is the one really doing the computation and the old one performWFlownStep 
https://github.com/qcdcode/quda/blob/5028c3a8a3ea0655307339893084633de11c76ee/lib/interface_quda.cpp#L5597
is initializing a structure param and calling the other function performWFlownStep_param, so that we do not lose the backward compatibility 

should we allocate the memory as we did in tmLQCD?
```
QudaGaugeObservableParam *param;
  param = (QudaGaugeObservableParam*) malloc(sizeof(QudaGaugeObservableParam) * (n_steps+1));
  for (int i=0;i<n_steps+1; i++){
    param[i] = newQudaGaugeObservableParam();   
    param[i].compute_plaquette = QUDA_BOOLEAN_TRUE;
    param[i].compute_qcharge = QUDA_BOOLEAN_TRUE; 
  }
  ...
  free(param)
```

> Ah, wait, I understand, sorry. Let's try it like this. I think, however, that `newQudaGaugeParam` should be called on all elements to make sure that any future changes don't break the initialisation.

We add the initialization of the params with newQudaGaugeParam() in a following commit.
https://github.com/qcdcode/quda/blob/5028c3a8a3ea0655307339893084633de11c76ee/lib/interface_quda.cpp#L5602

kostrzewa:
> Since we could not implement a dofoult value for a parameter in c we add a function in quda performWFlownStep_param that is the one really doing the computation and the old one performWFlownStep

Yes, I agree. I also agree with the `new` and `delete` pairing. I had misinterpreted your intent.

Having said that, the gradient flow (and other gauge smearing functions) in QUDA are quite odd in that they perform the measurements only as informative output. I think your miniminal modification proposal is a good way to see if Kate would be willing to accept further changes to make the output independent of the verbosity setting.

kostrzewa:
Does this now produce the same output as tmLQCD's gradient flow? (up to the difference in the topological charge at small flow times discussed in https://github.com/lattice/quda/issues/959)

Marcogarofalo:
> Does this now produce the same output as tmLQCD's gradient flow? (up to the difference in the topological charge at small flow times discussed in [lattice/quda#959](https://github.com/lattice/quda/issues/959))

Yes, also there is the addition of the QUDA output depending on its verbosity.

kostrzewa:
Perfect, thank you. I'll merge this in once the corresponding commit lands in QUDA. Alternatively, you can also add some guards to enable/disable this at the configuration stage so being able to compile does not depend on the QUDA-PR being merged.

Marcogarofalo:
I did not find I way to get the current verbosity of quda, that is why I add the function `_setVerbosityQuda()`. 
@kostrzewa @simone-romiti  do you have something better?

simone-romiti:
Maybe I misunderstood, but in ```quda/lib/util_quda.cpp``` there are the setVerbosity() and getVerbosity() functions. Is that what you're looking for?

kostrzewa:
There's unfortunately no corresponding C interface function for `getVerbosity()`. Maybe this is something that we actually need to change in QUDA.

However, looking at our `quda_interface.c`, we set the general verbosity to `QUDA_SUMMARIZE` independently of the debug level. At this stage then, I think you can do what you do, although I would explicitly set `QUDA_SILENT` before the gradient flow if `g_debug_level <= 2` and leave it at its default in all other cases.

kostrzewa:
It appears to be a regression in QUDA. I'll `git-bisect` and submit an issue.

kostrzewa:
Works now thanks to dbc1e288315201f35f0069c60e4f936daad54d22 in QUDA.

PR #518, opened 2022-01-21
expose 'QUDA_ENABLE_[DEVICE,PINNED]_MEMORY_POOL' via new parameters '…

kostrzewa:
…EnableDeviceMemoryPool' and 'EnablePinnedMemoryPool' and disable these by default to save device / host memory


kostrzewa:
@Marcogarofalo @pittlerf @sbacchio  FYI

PR #519, opened 2022-01-28
Quda work add actions bug measurements

simone-romiti:
quda was not working with the following input (2kappaMu not specified)

```
BeginOperator CLOVER
  CSW = 1.76
  kappa = 0.15
  SolverPrecision = 1e-14
  MaxSolverIterations = 1000
  solver = mg
  UseEvenOdd = yes
  useexternalinverter = quda
  usesloppyprecision = single  
EndOperator

BeginMeasurement CORRELATORS
  MaxSolverIterations = 1000
  Frequency = 1
EndMeasurement
```

### DIFF
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -1003,25 +1003,14 @@ int invert_eo_quda(spinor * const Even_new, spinor * const Odd_new,
                    SloppyPrecision sloppy_precision,
                    CompressionType compression) {
 
-  if(g_proc_id == 0) {
-    printf("hey1");
-  }
 
   tm_stopwatch_push(&g_timers, __func__, "");
 
   spinor ** solver_field = NULL;
   const int nr_sf = 2;
   init_solver_field(&solver_field, VOLUME, nr_sf);
-  if(g_proc_id == 0) {
-    printf("hey2");
-  }
 
   convert_eo_to_lexic(solver_field[0],  Even, Odd);
-
-  if(g_proc_id == 0) {
-    printf("hey3");
-  }
-
 
 // this is basically not necessary, but if we want to use an a nitial guess, it will be
 //  convert_eo_to_lexic(solver_field[1], Even_new, Odd_new);
@@ -1036,31 +1025,14 @@ int invert_eo_quda(spinor * const Even_new, spinor * const Odd_new,
 
   inv_param.kappa = g_kappa;
 
-  if(g_proc_id == 0) {
-    printf("hey4");
-  }
-
   // figure out which BC to use (theta, trivial...)
   set_boundary_conditions(&compression, &gauge_param);
   // set the sloppy precision of the mixed prec solver
 
-  if(g_proc_id == 0) {
-    printf("hey5");
-  }
-
   set_sloppy_prec(sloppy_precision, &gauge_param, &inv_param);
   
-    if(g_proc_id == 0) {
-    printf("hey6");
-  }
-
   // load gauge after setting precision
   _loadGaugeQuda(compression);
-
-  if(g_proc_id == 0) {
-    printf("hey7");
-  }
-
 
   // this will also construct the clover field and its inverse, if required
   // it will also run the MG setup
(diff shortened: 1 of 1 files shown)


quda_interface.c
```diff
@@ -1353,7 +1424,7 @@ void _setOneFlavourSolverParam(const double kappa, const double c_sw, const doub
     fatal_error("Attempt to set inv_param.tm_rho but --enable-quda_experimental was not set!", __func__);
 #endif
     } else {
-      inv_param.twist_flavor = QUDA_TWIST_NO;
+      inv_param.twist_flavor = QUDA_TWIST_SINGLET;
```
kostrzewa, line 1427: you have to be careful here: `QUDA_TWIST_NO` is correct when doing pure Wilson clover fermions (without mass preconditioning).

quda_interface.c
```diff
@@ -1060,11 +1111,31 @@ int invert_eo_quda(spinor * const Even_new, spinor * const Odd_new,
...
   int iteration = inv_param.iter;
 
+  if(g_proc_id == 0) {
+    printf("hey12");
+  }
+
+
   reorder_spinor_fromQuda( (double*)spinorOut, inv_param.cpu_prec, 0 );
+
+  if(g_proc_id == 0) {
```
kostrzewa, line 1121: These should be removed once you're happy.

kostrzewa (changes requested):
Thanks. Can you please make sure that the change that you're suggesting is appropriate also for the case of pure Wilson clover fermions? We shouldn't be adding a twist when none is required.

The debug statements should also be removed (the `heyXYZ` ones)

kostrzewa:
Note that operator `CLOVER` with `2kappamu = 0.0` implies `QUDA_TWIST_NO` for the QUDA operator.

Marcogarofalo:
with `QUDA_TWIST_NO` quda is complaining 
```
MG level 0 (GPU): ERROR: twist flavors do not match: 1 0 (rank 0, host lnode15.cluster.hiskp, color_spinor_field.cpp:707 in checkField())
MG level 0 (GPU):        last kernel called was (name=N4quda15CopyColorSpinorILi4ELi3ENS_11colorspinor11FloatNOrderIfLi4ELi3ELi4ELb0ELb0EEENS2_IdLi4ELi3ELi2ELb0ELb0EEESt5tupleIJRNS_16ColorSpinorFieldERKS6_19QudaFieldLocation_sPfPKdEEEE,volume=4x8x8x8,aux=GPU-offline,vol=2048,precision=8,order=2,Ns=4,Nc=3vol=2048,precision=4,order=4,Ns=4,Nc=3,PreserveBasis)
```
instead with `QUDA_TWIST_SINGLET` is working fine and the online measurements are the same of the host version.

Probably QUDA  is complaining because it is using an old MG setup with mu>0

kostrzewa:
> with QUDA_TWIST_NO quda is complaining
> Probably QUDA is complaining because it is using an old MG setup with mu>0

okay, but then we need to separate the cases:

1) pure Wilson clover HMC with Wilson clover online measurements
2) twisted mass clover HMC with twisted mass clover measurements

Any other combination does not make sense, although it is possible from the point of view of parameters.

Ideally, the MG setup would be forced to be reset when the operator type changes between HMC and online measurement (i.e., when we're running twisted clover HMC and perform a Wilson clover measurement or vice versa).

What I observed, however, is that twisted clover HMC + twisted clover online measurements don't work with the MG, is that correct?

Marcogarofalo:
> What I observed, however, is that twisted clover HMC + twisted clover online measurements don't work with the MG, is that correct?

I did not see any problem with this combination, more specific
```
BeginOperator CLOVER
  CSW = 1.76
  kappa = 0.15
  2kappamu = 0.0015846837
  SolverPrecision = 1e-14
  MaxSolverIterations = 1000
  solver = mg
  UseEvenOdd = yes
  useexternalinverter = quda
  usesloppyprecision = single
EndOperator

BeginMeasurement CORRELATORS
  MaxSolverIterations = 1000
  Frequency = 1
EndMeasurement



BeginMonomial CLOVERDET
  Timescale = 1
  kappa = 0.15
  2KappaMu = 0.0015846837
  CSW = 1.74
  rho = 0.09353509
  MaxSolverIterations = 1000
  AcceptancePrecision =  1.e-19
  ForcePrecision = 1.e-15
  Name = cloverdetlight
  solver = mg
  useexternalinverter = quda
  usesloppyprecision = single
EndMonomial

BeginMonomial CLOVERDETRATIO
  Timescale = 1
  kappa = 0.15
  2KappaMu = 0.0015846837
  rho = 0.01039279
  rho2 = 0.09353509
  CSW = 1.74
  MaxSolverIterations = 1000
  AcceptancePrecision =  1.e-19
  ForcePrecision = 1.e-16
  Name = cloverdetratio1light
  solver = mg
  useexternalinverter = quda
  usesloppyprecision = single
EndMonomial

```

kostrzewa:
> I did not see any problem with this combination, more specific

The residual check after the online measurement comes out correctly?

```
[...]
# Inversion done in N iterations, squared residue = 3.419632e+04!
# Inversion done in X sec. 
# : Time for correlators_measurement X s level: 1 proc_id: 0 /HMC/correlators_measurement
[...]
```

kostrzewa:
About your input file: make sure that you have the same value of csw everywhere. I think the setup will always be rebuilt when you don't.

Issue #517, opened 2022-01-11
MonitorForces not working with QUDA


sbacchio:
When `MonitorForces = yes` an error occurs in QUDA. More investigation required.

The error is the following:
```
MG level 0 (GPU): ERROR: Spinor volume 351232 doesn't match gauge volume 0 (rank 0, host jwb0134.juwels, dirac.cpp:125 in checkParitySpinor())
MG level 0 (GPU):        last kernel called was (name=N4quda4blas7axpbyz_IfEE,volume=28x56x56x4,aux=GPU-offline,vol=351232,precision=4,order=4,Ns=4,Nc=3,TwistFlavour=1)
```

For more details see `$SCRATCH_fssh/bacchio1/C56/logs/log_trial_4811972.out` on Juwels Booster

kostrzewa:
We've reproduced this, albeit with a different error: we get a precision mismatch. This indicates that one of the parameter structs is not properly initialized.

pittlerf:
Hi, yes, in case I do 
in the beginning of monitor_force-s 
`  update_tm_gauge_id(&g_gauge_state, 0.1);` 
and
`  update_tm_gauge_id(&g_gauge_state, -0.1);
`
the problem actually disappers

pittlerf:
> We've reproduced this, albeit with a different error: we get a precision mismatch. This indicates that one of the parameter structs is not properly initialized (or overwritten somehow).

I saw similar kind of issue using hot start:
like
`MG level 0 (GPU): ERROR: Precisions 4 8 do not match (/cyclamen/home/fpittler/code/quda_ndeg/lib/../include/kernels/dslash_wilson.cuh:51 in WilsonArg())
`

sunpho84:
Hi
I'm seeing a similar error, I attach the relevant part of the valgrind inspection. My understanding is that the check which control whether the sloppy gauge must be allocated, 

https://github.com/sunpho84/quda/blob/5431b168b09343503d0d676425069dc895879c92/lib/interface_quda.cpp#L670-L674

is not working, I have to say I don't understand the logic.

Below is also the relevant part of the input file, maybe somebody can spot a parameter not properly set?


```==93804== Invalid read of size 8
==93804==    at 0x14932B84: quda::Dirac::checkParitySpinor(quda::ColorSpinorField const&, quda::ColorSpinorField const&) const (dirac.cpp:122)
[...]

PR #525, opened 2022-02-17
partial workaround for some of the precision mismatch problems in QUDA interface

kostrzewa:
We were not tracking the precisions of the gauge and clover fields present on the device and hence these were bound to lead to mismatches when switching from one monomial employing an operator and solver with a particular set of precisions to another monomial employing a different set of precisions (for `cuda_prec`, `cuda_prec_sloppy`, `cuda_prec_refinement_sloppy`, `cuda_prec_precondition` and `cuda_prec_eigensolver` and the corresponding ones for `clover_quda_*`.

Unfortunately, this does not fully "solve" #517, but it does raise a new type of error there which might help resolve that too in the end.

The set of changes here has a drawback of course: the gauge and clover fields are reloaded much more frequently instead of just causing the missing precision to be instantiated from the existing double precision field on the device. Not sure how bad the additional overhead is compared to the time spent in a trajectory.

It is *especially* a complete waste of time to call `reorder_gauge_toQuda` so frequently because this should really only be called when `g_gauge_field` or any of the theta angles have *actually* changed. `freeGaugeQuda()` and `loadGaugeQuda(..)` do have to be called, however (at least for now, since that is our mechanism for ensuring that the field is up to date).


kostrzewa:
Do not merge yet. I'm afraid there are other issues left...

kostrzewa:
https://github.com/etmc/tmLQCD/pull/525/commits/6c040df7fb9c312d6adcabf672db9e031150bbd8 resolves #517 and solidifies the precision mismatch fix

The problem was that the MG Setup (in particular I guess the coarse operators) seem to have an internal memory of the gauge field device pointers (rather than an abstract reference which I would expect to update with the gauge field on the device).

When we call `freeGaugeQuda()` in the HMC, we are left with dangling pointers in the MG and this is what causes the volume mismatches. At the same time, the current gauge and clover fields must be consistent with the precisions in the MG Setup and this leads to the precision mismatches.

I'm not happy with this because it induces lots of MG Setup updates, but these are not THAT expensive. I think this is ready to test now.

Marcogarofalo:
> I'm not happy with this because it induces lots of MG Setup updates, but these are not THAT expensive. I think this is ready to test now.

but it works. Also all the valgrind messages disappear. Thanks

kostrzewa:
> but it works. Also all the valgrind messages disappear. Thanks

Thanks for the test! It would be interesting to see how a profile with this code compares to the profiles that you generated a while ago.

kostrzewa:
Thanks for all the tests, valgrind runs and tentative workarounds @sunpho84 @simone-romiti @Marcogarofalo @pittlerf  Without the hints from these I would not have been able to fix this...

PR #528, opened 2022-02-23
fix online meaurements with QUDA when running the HMC

kostrzewa:
always set inv_param.dagger to make sure that we don't accidentally re-use the parameter from the previous solve

### DIFF
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -981,6 +981,10 @@ int invert_quda_direct(double * const propagator, double const * const source,
                             optr->maxiter,
                             0, 0);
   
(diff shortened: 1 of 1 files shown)


kostrzewa:
Would appreciate a test of this.

Marcogarofalo:
It works 
```
/qbigwork/garofalo/test_tmLQCD_QUDA/online_meas$ grep residue slurm-86651.out
# Inversion done in 27 iterations, squared residue = 3.327151e-14!
# Inversion done in 24 iterations, squared residue = 2.201307e-14!
```

Marcogarofalo (approved):

Issue #526, opened 2022-02-23
abort HMC if any monomial_solve reports -1 iterations


kostrzewa:
add global parameter which controls if HMC should be aborted in case of non-convergence

PR #531, opened 2022-03-03
Check monomials convergence

simone-romiti:
Reply to the issue https://github.com/etmc/tmLQCD/issues/526 .
I added a global parameter which is specified in the input file as "yes" or "no". The default is "no". 
If set to "yes" the convergence is checked for each monomial. If the returned number of iterations equals -1 , the program is aborted.


kostrzewa:
Thanks! (note that I've changed the target branch). Could you also add a description of this option to the TeX-documentation?

simone-romiti:
Done! Please check if the explanation is clear enough.

PR #533, opened 2022-03-07
treat setup updates and refreshes differently from the point of view …

### LINKED ISSUE #532: HMC setup refresh never reached since b70e6f40b441bde1c0b3eff108f5ce58312e1221
It seems that b70e6f40b441bde1c0b3eff108f5ce58312e1221 has introduced a regression into the setup refresh logic. Instead of refreshing when `MGRefreshSetupMDUThreshold` is reached, the setup is merely updated. This of course destroys convergence as the trajectory progresses. Thanks @Marcogarofalo for the report.

kostrzewa:
tm_QudaMGSetupState_t

This should fix #532

### DIFF
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -1325,9 +1325,7 @@ void _setOneFlavourSolverParam(const double kappa, const double c_sw, const doub
   inv_param.tol = sqrt(eps_sq);
   inv_param.maxiter = maxiter;
   inv_param.Ls = 1;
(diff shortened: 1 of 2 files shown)


Marcogarofalo:
Thank you, I am trying it on marconi100.

Marcogarofalo (approved):
Now, cA211.12.48 runs without problems. Thank you for the fix

PR #534, opened 2022-03-09
tables position within same section + caption

Marcogarofalo:
I found confusing in the profile tool the position of the tables
example_table_out_section

so I did two things:
- Force the table to be placed in the same subsection;
- Add a caption to the table.

the result is now the following
image


kostrzewa:
looks good, thanks. I left it flexible because the size of the table can vary a lot, but I agree that this is better in any case

PR #535, opened 2022-03-10
print message only if g_proc_id==0

Marcogarofalo:


kostrzewa:
thanks, I missed that!

simone-romiti:
Thank you Marco!

Issue #527, opened 2022-02-23
add mechanism to QUDA interface to regenerate setup if an MG-based solve  in the HMC does not converge


kostrzewa:

simone-romiti:
Please see pull request: https://github.com/etmc/tmLQCD/pull/536

PR #536, opened 2022-03-13
Quda regenerate mg setup

simone-romiti:
Reply to the issue https://github.com/etmc/tmLQCD/issues/527 .

### Edits

I edited the ```quda_interface.c``` file as follows:
- I created a function ```void forceRegenerateMGSetupQuda(const CompressionType compression)``` which regenerates the MG setup.
  * At the beginning ```_loadGaugeQuda()``` is called. For how it is implemented now, maybe it's redundant because this function is always called after one call of that function.
  * In order to destroy the MG setup, the same MG-related functions called in ```_endQuda()``` are used.
  * In order to regenerate the MG setup, the same MG-related functions called in ```_initQuda()``` are used.
- This function is called inside each ```invert_*()``` function defined in that file, whenever one's using the MG and the latter didn't converge.
- Any time the MG doesn't converge, the inversion is done one more time in order to check for possible instabilities. This is implemented through a global variable updated as the MG doesn't converge.

### Note

We can think about adding a global parameter, to be specified in the input file, which limits the number of regenerations of the MG setup. Now the latter is implicitly set to 1.

### DIFF
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -2182,9 +2182,8 @@ int invert_eo_degenerate_quda(spinor * const out,
 
     if (solver_flag == MG) {
       // re-running (once) the inversion which failed
-      if (quda_mg_setup_state->force_reset == 0) {
-        quda_mg_setup_state->force_reset = 1;
-        _updateQudaMultigridPreconditioner();  // regenerate the MG setup
+      if (quda_mg_setup_state.force_reset == 0) {
+        quda_mg_setup_state.force_reset = 1;
         return invert_eo_degenerate_quda(out, in, precision, max_iter, solver_flag, rel_prec,
                                          even_odd_flag, solver_params, sloppy_precision,
                                          compression, QpQm);
@@ -2195,10 +2194,6 @@ int invert_eo_degenerate_quda(spinor * const out,
       }
     }
     return (-1);
-  } else {
-    if (g_proc_id == 0) {
-      quda_mg_inversion_failed = 0; // resetting for the next solver call
-    }
   }
 
   return(iterations);
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -2239,6 +2239,10 @@ int invert_eo_degenerate_quda(spinor * const out,
       }
     }
     return (-1);
+  } else {
+    if (g_proc_id == 0) {
+      quda_mg_inversion_failed = 0; // resetting for the next solver call
+    }
   }
 
   return(iterations);
@@ -2340,6 +2344,10 @@ int invert_eo_quda_oneflavour_mshift(spinor ** const out,
       }
     }
     return(-1);
+  } else {
+    if (g_proc_id == 0) {
+      quda_mg_inversion_failed = 0; // resetting for the next solver call
+    }
   }
 
   return(iterations);
@@ -2470,6 +2478,10 @@ int invert_eo_quda_twoflavour_mshift(spinor ** const out_up, spinor ** const out
       }
     }
     return(-1);
+  } else {
+    if (g_proc_id == 0) {
+      quda_mg_inversion_failed = 0; // resetting for the next solver call
+    }
   }
 
   return(iterations);
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -2191,7 +2191,8 @@ int invert_eo_degenerate_quda(spinor * const out,
     if (solver_flag == MG) {
       // when the MG fails to converge, the reason may be a degenerated MG setup
       // or, if this happens in the HMC, bad luck with the evolved setup
-      // we try to regenerate it once and abort if that doesn't help
+      // we try to force refresh it once and abort if that doesn't help
+
       // we use this variable to break out of the recursion 
       static int quda_mg_setup_was_force_refreshed = 0;
       if( quda_mg_setup_was_force_refreshed == 0 ){
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -1543,15 +1543,7 @@ void _updateQudaMultigridPreconditioner(){
     }
     inv_param.preconditioner = quda_mg_preconditioner;
 
-    // during a force reset we only update the parameters tracked in the MG state
-    // this is in order to not disrupt the normal sequence of update and refresh 
-    // operations as a function of the gauge_id
-    if(quda_mg_setup_state.force_refresh == 1){
-      quda_mg_setup_state_update(&quda_mg_setup_state, &quda_gauge_state, &quda_clover_state,
-                                 g_mu, g_kappa, g_c_sw);
-    } else {
-      set_quda_mg_setup_state(&quda_mg_setup_state, &quda_gauge_state, &quda_clover_state);
-    }
+    set_quda_mg_setup_state(&quda_mg_setup_state, &quda_gauge_state, &quda_clover_state);
 
     tm_stopwatch_pop(&g_timers, 0, 1, "TM_QUDA");
 
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -144,12 +144,6 @@ void* quda_mg_preconditioner;
 // on the given level)
 QudaEigParam mg_eig_param[QUDA_MAX_MG_LEVEL];
 
-/**
- * When using the quda MG solver, this variable stores the number of failed inversions.
- * It is used in order to decide how many times to retry the inversion after regenerating the MG setup.
- */
-int quda_mg_inversion_failed = 0;
-
 // input params specific to tmLQCD QUDA interface
 tm_QudaParams_t quda_input;
 
@@ -190,14 +184,6 @@ void _setOneFlavourSolverParam(const double kappa, const double c_sw, const doub
                                const int solver_type, const int even_odd,
                                const double eps_sq, const int maxiter,
                                const int single_parity_solve, const int QpQm);
-
-/**
- * Regenerate the MG setup. To be used e.g. when an MG-based solver in doesn't converge.
- * @param mg_param Contains all metadata regarding host and device
-   *               storage and solver parameters
- */
-void forceRegenerateMGSetupQuda(const CompressionType compression);
-
 
 void set_default_gauge_param(QudaGaugeParam * gauge_param){
   // local lattice size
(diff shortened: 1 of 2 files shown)


Marcogarofalo:
Maybe you want to merge this to `quda_work_add_action` instead of `master` branch

simone-romiti:
> Maybe you want to merge this to `quda_work_add_action` instead of `master` branch

Yes, thanks

quda_interface.c
```diff
@@ -2254,8 +2321,34 @@ int invert_eo_quda_oneflavour_mshift(spinor ** const out,
 
   tm_stopwatch_pop(&g_timers, 0, 0, "TM_QUDA");
 
-  if(iterations >= max_iter)
+  if(iterations >= max_iter){
```
kostrzewa, line 2324: for now this is not required here as we don't use MG for multi-shift solves

quda_interface.c
```diff
@@ -2361,8 +2454,35 @@ int invert_eo_quda_twoflavour_mshift(spinor ** const out_up, spinor ** const out
...
```
kostrzewa, line 2457: this is not required here as we don't use the MG for multi-shift solves

quda_interface.c
```diff
@@ -2052,6 +2066,32 @@ void _setQudaMultigridParam(QudaMultigridParam* mg_param) {
   strcpy(mg_param->vec_outfile, "");
 }
 
+void forceRegenerateMGSetupQuda(const CompressionType compression){
```
kostrzewa, line 2069: This function is not necessary and duplicates things which are done elsewhere. It also resets all the parameters, so it does more than just regenerating the setup: it completely resets the state of all MG-related parameters in the QUDA interface.

Here's what I would do:

* add a `force_reset` flag to `tm_QudaMGSetupState_t`, which defaults to `0`, of course
* when an MG solve fails, `force_reset` is set to `1`
* modify `check_quda_mg_setup_state` to return `TM_QUDA_MG_SETUP_RESET` when `force_reset == 1`

This should be enough to force the setup to be regenerated when you call `invert_eo_degenerate_quda` recursively in the `quda_mg_inversion_failed` check.

kostrzewa (changes requested):
Thanks, I think this needs some work to make use of the `tm_QudaMGSetupState_t` and the existing infrastructure of generating the MG setup. The proposed changes also do more than just reset the MG setup when `forceRegenerateMGSetupQuda` is called in isolation.

It's not necessary to have this logic in the multi-shift solver interfaces since we don't use the MG solver there.

quda_interface.c
kostrzewa: > Note also that you only have to retry the solve once. A regenerated setup must work, otherwise the parameters are simply poor and the job should abort.

This also means that you don't really need the new global variable to track how many times the inversion using MG has failed. It's sufficient to catch the non-convergence and try again (once) right in `invert_eo_degenerate_quda`.

quda_interface.c
simone-romiti: ok, removed. I left thoe lines only in ```invert_eo_degenerate_quda()```

quda_interface.c
simone-romiti: ok, removed. I left those lines only in ```invert_eo_degenerate_quda()```

quda_types.h
```diff
@@ -126,6 +126,7 @@ typedef struct tm_QudaMGSetupState_t {
   double theta_y;
   double theta_z;
   double theta_t;
+  int force_reset = 0; // set to 1 when the MG doesn't converge
```
kostrzewa, line 129: you can't do this in C, the default value needs to be set in some function (as are all the other defaults)

quda_interface.c
```diff
@@ -2175,8 +2175,31 @@ int invert_eo_degenerate_quda(spinor * const out,
...
+  if (iterations >= max_iter) {
+    if (g_proc_id == 0) {
+      printf("# TM_QUDA: iterations == -1\n");
+    }
+
+    if (solver_flag == MG) {
+      // re-running (once) the inversion which failed
+      if (quda_mg_setup_state->force_reset == 0) {
+        quda_mg_setup_state->force_reset = 1;
+        _updateQudaMultigridPreconditioner();  // regenerate the MG setup
```
kostrzewa, line 2187: you don't need to call this yourself, the call to `invert_eo_degenerate_quda` will do it

simone-romiti: Thank you, I removed it

quda_types.h
simone-romiti: Yes, with the last commit I modified the ```set_quda_mg_setup_state()``` function such that ```force_reset``` is set to ```0```.

quda_interface.c
```diff
@@ -2175,8 +2175,26 @@ int invert_eo_degenerate_quda(spinor * const out,
...
+  if (iterations >= max_iter) {
+    if (g_proc_id == 0) {
+      printf("# TM_QUDA: iterations == -1\n");
+    }
+
+    if (solver_flag == MG) {
+      // re-running (once) the inversion which failed
+      if (quda_mg_setup_state.force_reset == 0) {
+        quda_mg_setup_state.force_reset = 1;
+        return invert_eo_degenerate_quda(out, in, precision, max_iter, solver_flag, rel_prec,
```
kostrzewa, line 2187: just store the return value of `invert_eo_degenerate_quda` here and if it's equal to -1, you can call `fatal_error` right here

that way you don't need to check `quda_mg_setup_state.force_reset == 0` here

kostrzewa:
Thanks. I've cleaned things up a bit and added a local static variable to break out of the recursion. This seems to work as expected now. I've explicitly let an MG setup deteriorate by not evolving it with the gauge field and as the number of iterations begins to increase, the mechanism is triggered. The first time around, it is able to save the run. At the second occurence it fails and the program terminates:

```
GCR: Convergence at 48 iterations, L2 relative residual: iterated = 2.507958e-10, true = 2.507958e-10 (requested = 3.162278e-10)
GCR: Convergence at 52 iterations, L2 relative residual: iterated = 2.348466e-10, true = 2.348466e-10 (requested = 3.162278e-10)
GCR: Convergence at 52 iterations, L2 relative residual: iterated = 2.990532e-10, true = 2.990532e-10 (requested = 3.162278e-10)
GCR: Convergence at 59 iterations, L2 relative residual: iterated = 2.469045e-10, true = 2.469045e-10 (requested = 3.162278e-10)
GCR: Convergence at 67 iterations, L2 relative residual: iterated = 2.912129e-10, true = 2.912129e-10 (requested = 3.162278e-10)
GCR: Convergence at 75 iterations, L2 relative residual: iterated = 2.611657e-10, true = 2.611657e-10 (requested = 3.162278e-10)
GCR: Convergence at 96 iterations, L2 relative residual: iterated = 2.987194e-10, true = 2.987194e-10 (requested = 3.162278e-10)
GCR: Convergence at 110 iterations, L2 relative residual: iterated = 2.938478e-10, true = 2.938478e-10 (requested = 3.162278e-10)
GCR: Convergence at 147 iterations, L2 relative residual: iterated = 3.078089e-10, true = 3.078089e-10 (requested = 3.162278e-10)
GCR: Convergence at 169 iterations, L2 relative residual: iterated = 2.863158e-10, true = 2.863158e-10 (requested = 3.162278e-10)
GCR: Convergence at 126 iterations, L2 relative residual: iterated = 3.055091e-10, true = 3.055091e-10 (requested = 3.162278e-10)
GCR: Convergence at 151 iterations, L2 relative residual: iterated = 2.725082e-10, true = 2.725082e-10 (requested = 3.162278e-10)
GCR: Convergence at 161 iterations, L2 relative residual: iterated = 2.883943e-10, true = 2.883943e-10 (requested = 3.162278e-10)
GCR: Convergence at 207 iterations, L2 relative residual: iterated = 3.118657e-10, true = 3.118657e-10 (requested = 3.162278e-10)
### -> setup is regenerated here
GCR: Convergence at 34 iterations, L2 relative residual: iterated = 2.830836e-10, true = 2.830836e-10 (requested = 3.162278e-10)
GCR: Convergence at 37 iterations, L2 relative residual: iterated = 1.898408e-10, true = 1.898408e-10 (requested = 3.162278e-10)
GCR: Convergence at 28 iterations, L2 relative residual: iterated = 2.033467e-10, true = 2.033467e-10 (requested = 3.162278e-10)
GCR: Convergence at 29 iterations, L2 relative residual: iterated = 2.099869e-10, true = 2.099869e-10 (requested = 3.162278e-10)
### -> no convergence possible any more
GCR: Convergence at 300 iterations, L2 relative residual: iterated = 3.994108e-02, true = 3.994108e-02 (requested = 3.162278e-10)
GCR: Convergence at 300 iterations, L2 relative residual: iterated = 6.097005e-01, true = 6.097005e-01 (requested = 3.162278e-10)
GCR: Convergence at 300 iterations, L2 relative residual: iterated = 1.555925e-01, true = 1.555925e-01 (requested = 3.162278e-10)
GCR: Convergence at 300 iterations, L2 relative residual: iterated = 8.021428e-01, true = 8.021428e-01 (requested = 3.162278e-10)
### program terminates
```

The fact that the solves just after the setup regeneration work and the next four solves fail suggests to me that there is still a bit of a logic problem present. To be fair, however, this was a run with two monomials using the MG and we know that it's problematic when these have very different `rho` values.

kostrzewa:
Okay, I fixed the issue that existed with at the end of my test there.

I've also replaced *resetting* the setup with *refreshing* it instead, which should be sufficient to make the MG work properly when it deteriorates. We'll see. Refreshing is much faster than resetting, of course.

kostrzewa:
@simone-romiti will you be able to run some tests of this? I don't know if I caught all possible ways this can go wrong...

simone-romiti:
I've tested your solution and it works as expected also for me.
For reference, I did it with the help of _ad hoc_ ```static int``` variables here and there as follows:
* I forced the non-convergence editing the line https://github.com/etmc/tmLQCD/blob/b061909a7afd24dd19188d418d9c59f92d67c087/quda_interface.c#L2190 in order to check that the regeneration of the MG is done only once. Namely, the program doesn't enter this scope twice: https://github.com/etmc/tmLQCD/blob/b061909a7afd24dd19188d418d9c59f92d67c087/quda_interface.c#L2198 .
* I did also a separate test where I forced the non-convergence after the first failure of the MG and consequent regeneration of the MG, editing the line https://github.com/etmc/tmLQCD/blob/b061909a7afd24dd19188d418d9c59f92d67c087/quda_interface.c#L2208 . The program aborts correctly.

PR #538, opened 2022-04-06
Fixed overflow in printout of rectan-term to output.data (appears for…

Finkenrath:
Normalisation factor for rectangular term is not converted before division. Leads to an overflow, when lattice is of size 112x112x112x224.


kostrzewa (approved):

kostrzewa:
thanks!

PR #541, opened 2022-05-19
specify quda backend

simone-romiti:
in progress: changes to configure.in so that we can support both cuda and hip backends for quda.


kostrzewa:
I merged #542 after being told:

> I've done a test linking with CUDA on marconi100 and run a successful job.

At what time does it complain about `libcublas` missing?

kostrzewa:
note that it complains about `libblas` and not `libcublas`

PR #543, opened 2022-05-27
remove legacy GPU support

kostrzewa:
The presence of both legacy GPU code and the QUDA interface is confusing, especially since the former is completely unmaintained. This PR removes the legacy code completely.


kostrzewa:
@Marcogarofalo @simone-romiti I think I got all remnants of the legacy GPU code, it would be great if you could take a look, however.

@urbach Do you agree with the complete removal of the legacy GPU code?

urbach:
> @Marcogarofalo @simone-romiti I think I got all remnants of the legacy GPU code, it would be great if you could take a look, however.
> 
> @urbach Do you agree with the complete removal of the legacy GPU code?

definitly!

kostrzewa:
@Marcogarofalo @simone-romiti did you have a chance yet to check if I've forgotten to remove something?

Marcogarofalo:
I checked this version and it works. Maybe there is still some legacy file or line that can be eliminated in the future. I will suggest proceeding with the merge.

PR #546, opened 2022-07-08
Hb solver

kostrzewa:


read_input.l
```diff
@@ -2418,6 +2419,57 @@ static inline double fltlist_next_token(int * const list_end){
   }
 }
 
+<DETMONOMIAL,CLDETMONOMIAL,CLDETRATMONOMIAL,CLDETRATRWMONOMIAL,NDRATMONOMIAL,NDRATCORMONOMIAL,NDCLRATMONOMIAL,NDCLRATCORMONOMIAL,RATMONOMIAL,RATCORMONOMIAL,CLRATMONOMIAL,CLRATCORMONOMIAL,NDDETRATMONOMIAL,NDCLDETRATMONOMIAL>{
+  {SPC}*UseExternalInverter{EQL}quda {
```
kostrzewa, line 2423: The parsing strings need to be renamed too, otherwise the parse cannot pick which one it will call.. Something like `HBUseExternalInverter`

read_input.l
```diff
@@ -2446,6 +2498,10 @@ static inline double fltlist_next_token(int * const list_end){
    solver_caller=YY_START;
    BEGIN(NDMSOLVER);
   }
+  {SPC}*HB_Solver{EQL} {
+   solver_caller=YY_START;
+   BEGIN(NDMSOLVER);
```
kostrzewa, line 2503: are you sure this is correct? Do we need to be able to set the HB solver also for the ND monomial?

kostrzewa:
@simone-romiti opened a pull request to review this, please see the inline comments

monomial/monomial.c
```diff
@@ -248,6 +250,11 @@ int init_monomials(const int V, const int even_odd_flag) {
         }
       }
       else if(monomial_list[i].type == CLOVERDETRATIO) {
+        if (monomial_list[i].HB_solver == _default_HB_solver_flag) {
+          monomial_list[i].HB_solver = monomial_list[i].solver;
+        }
+        monomial_list[i].HB_solver_params = monomial_list[i].solver_params;
```
Marcogarofalo, line 256: please check this. As far as I can tell `HB_solver_params` need always to be initialised , so I moved outside the if guard.

Using the monomial 
```
BeginMonomial CLOVERDETRATIO
  Timescale = 0
  kappa = 0.1400645
  2KappaMu = 0.000224103200
  # numerator shift
  rho = 0.0016
  # denominator shift, should match CLOVERDET shift
  rho2 = 0.0128
  CSW = 1.74
  AcceptancePrecision =  1.e-21
  ForcePrecision = 1.e-18
  Name = cloverdetratio2light
  useexternalinverter = quda
  MaxSolverIterations = 500
  solver = mg
  usesloppyprecision = single
  # HB_solver = cg
  # HB_usesloppyprecision = single
  # HB_UseExternalInverter = quda
  # HB_MaxSolverIterations = 50000
EndMonomial
```
I got the same onlinemeas after one trajectory:

- commenting out he HB_solver lines
```
==> HB_solver/onlinemeas.000001 <==
6  1  30  2.754953e-30  3.374063e-31
6  1  31  3.400805e-32  -2.501565e-31
6  1  32  -2.318780e-33  0.000000e+00
```
- keeping only the comment in the line `# HB_UseExternalInverter = quda` i.e. using the  tmLQCD cg

```
==> HB_solver/onlinemeas.000001_with_HB_CPU_cg <==
6  1  30  2.754159e-30  3.371860e-31
6  1  31  3.399023e-32  -2.501224e-31
6  1  32  -2.242958e-33  0.000000e+00
```
- keeping `HB_solver` commented out
```
==> HB_solver_same_input/onlinemeas.000001 <==
6  1  30  2.754809e-30  3.377900e-31
6  1  31  3.415679e-32  -2.501734e-31
6  1  32  -2.284831e-33  0.000000e+00
```
-  using 6b31350 ad of course `HB_solver` commented out
```
==> add_actions/onlinemeas.000001 <==
6  1  30  2.755184e-30  3.367085e-31
6  1  31  3.416626e-32  -2.500371e-31
6  1  32  -2.259483e-33  0.000000e+00
```
these these were done in 
`/qbigwork/garofalo/test_tmLQCD_QUDA/cA211.08.32`

tmLQCD.kdev4
```diff
@@ -0,0 +1,4 @@
+[Project]
```
kostrzewa, line 1: please delete

workspace.code-workspace
```diff
@@ -0,0 +1,17 @@
+{
```
kostrzewa, line 1: please delete

sample-input/sample-hmc-tmcloverdet-GPU.input
```diff
@@ -0,0 +1,75 @@
+# this is a sample input file for a single cloverdet
```
kostrzewa, line 1: this should not be here, I thought I removed this with the GPU removal commit

monomial/monomial.c
kostrzewa: well, there should be defaults set for a few members in `HB_solver_params` as is done for `solver_params`. One way of doing this might indeed be to simply copy `solver_params`, but only if this does not lead to things being overwritten.

kostrzewa: I think this will overwrite settings. Any defaults should be set in `add_monomial`, which is called while reading the input file. `init_monomials`, instead, is called *after* the input file has already been parsed.

kostrzewa: A good way would be to set `mnl_list[i].HB_solver_params = mnl_list[i].solver_params` in `add_monomial` after the corresponding defaults in `solver_params` have been set

sample-input/sample-hmc-tmcloverdet-GPU.input
Marcogarofalo: sorry I made, some confusion with the directory.

read_input.l
```diff
@@ -2717,6 +2779,51 @@ static inline double fltlist_next_token(int * const list_end){
   }
 }
 
+<HB_MSOLVER>{
+  CG {
+    if(myverbose) printf("  HB Solver set to \"%s\" line %d monomial %d\n", yytext, line_of_file, current_monomial);
+    mnl->HB_solver = CG;
+    BEGIN(solver_caller);
```
Marcogarofalo, line 2786: Please check this line. We found that `BEGIN(solver_caller)` is necessary to continue reading the input file correctly but I do not understand the logic.

Marcogarofalo:
regarding the performance:

- mg heatbath
image

- With the cg in cloverdetratio2light heatbath we move the time to setup the mg in the cloverdetratio3light
image

- With the cg in cloverdetratio2light and cloverdetratio3light heatbath 
image

There is a slight improvement overall but mainly we are moving the time for `MG_Preconditioner_Setup` between monomials

kostrzewa:
Can you provide some more details on where these numbers were obtained and what is the ensemble in question?

kostrzewa:
The only problematic monomial is the heatbath of cloverdetratio2light (at least with our current choice of mass preconditioning). You should be looking at the solve time rather than setup+solve as the setup has to be done anyway (as you say). 

The problem is most acute on M100, where the solve for in the cloverdetratio2light takes several hundred seconds and much less so on other machines. Also don't forget that the setup is only run once per job (when acceptance is sufficiently high).

See the cA211.08.64 run:

```
19:51 bkostrze@login02 /m100_work/INF22_lqcd123_0/romiti/runs/cA211.08.64_therm/jobscript/logs 
 $ grep "Time for cloverdetratio_heatbath" log_cA211.08.64_therm_7288769_4294967294.out | grep ratio2
# : Time for cloverdetratio_heatbath 1.062495e+03 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 6.671943e+02 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
```

First call is with the setup, the second call without. On a 32c64 test lattice (or another test lattice), the problem will not be this massive, of course. Similary, on Juwels Booster the MG performs much better, so even if it does 300-400 iterations, the impact on runtime is not too bad (using CG would still be faster though).

Marcogarofalo:
Sorry I mean that test was on the same ensamble of the previous comment
```
/qbigwork/garofalo/test_tmLQCD_QUDA/cA211.08.32
```
There is a subdirectory add_actions which is my reference version with mg hb, please look at the last log file.

For the cg hb there is a directory HB_solver and i was quoting the second last log file(cg hb only in cloverdet2light) and the last log file with cg hb also in cloverdet3light.

read_input.l
kostrzewa: Indeed. This allows the parser to return to the point it came from (after having parsed the solver string). To be precise, this is what happens in flex:

```
<CLDETRATMONOMIAL,DETRATMONOMIAL>{
  {SPC}*HB_Solver{EQL} {
   solver_caller=YY_START;
   BEGIN(HB_MSOLVER);  -> jump to HB_MSOLVER parser
  }
  [...] -->
  <HB_MSOLVER>{
  CG {
    if(myverbose) printf("  HB Solver set to \"%s\" line %d monomial %d\n", yytext, line_of_file, current_monomial);
    mnl->HB_solver = CG;
    BEGIN(solver_caller);  --> jump back
  }  
  [...]
  } <-- return to this point
    {SPC}*HB_MaxSolverIterations{EQL}{DIGIT}+ {
    sscanf(yytext, " %[a-zA-Z_] = %d", name, &a);
    mnl->HB_maxiter = a;
    if(myverbose) printf("  HB_MaxSolverIterations set to %d line %d monomial %d\n", a, line_of_file, current_monomial);
  }
  [...]
```

While looking at this I realised that there are actually a few bugs in the parser. There are instances of `name_caller` where instead there should be instances of `solver_caller`...

kostrzewa: I'm actually surprised that the parsing of the monomials works correctly when the MG (or mixedbicgstab) is used as a solver... :)

kostrzewa:
@simone-romiti Did you have a chance to test how this behaves on M100?

kostrzewa:
I've just brought this up to date with `quda_work`.

kostrzewa:
TODO

- [ ] documentation
- [ ] performance improvement test on Marconi 100

kostrzewa:
@simone-romiti could you please finish the documentation for this so we can merge it?

doc/input.tex
```diff
@@ -453,16 +453,25 @@ \subsection{Input parameter for main program}
...
   \item {\ttfamily Kappa2}
-  \item {\ttfamily Name}: a name to be assigned to the monomial. The
-    default is {\ttfamily DETRATIO}
+  \item {\ttfamily Name}: 
+    a name to be assigned to the monomial. 
+    The default is {\ttfamily DETRATIO}
+  \item {\ttfamily HB\_solver}: 
+    Optional parameter. It's the solver to be used in the heatbath. 
+    If not specified, the MD solver is used.    
+    Supported values: MG, BICGSTAB.
```
kostrzewa, line 468: this supports many more values, not sure if it makes so much sense to list them here or whether we should introduce a new section into the documentation describing which solvers can be used in the HMC under which circumstances and with which kinds of fermions

simone-romiti: I see what you mean. I listed them here because at the moment the code foresees only these options. For these few lines of documentation I thought it was unnecessary to create a new section of this, which may be done when the number of tested cases increases. I don't have strong preference though.

kostrzewa:
Please don't forget to also document the other `HB_` parameters beyond just `HB_solver`.

doc/input.tex
kostrzewa: The fact that we have 

```
if( mnl->HB_solver == MG || mnl->HB_solver == BICGSTAB ){
```

there is because when these solvers are used, a different kind of solve is performed.

kostrzewa: Both MG and BICGSTAB can directly invert Q_{+}, rather than having to invert the normal operator Q_{+} Q_{-} and subsequently multiplying the result by Q_{-}. For BICGSTAB, this statement does not really apply for twisted mass fermions (because the solver does not perform very well in this case), but it certainly works for Wilson (clover) fermions without a twisted mass term. MG instead can deal with any of Wilson (clover) (twisted mass).

kostrzewa:
Thanks. Can you also check what's going on with the integration tests?

simone-romiti:
> Thanks. Can you also check what's going on with the integration tests?

On meluxina I did a test here:
`/mnt/tier1/project/p200094/romiti/tmLQCD_runs/iwa_1.745-csw1.7112-L32/test_HB_solver`.
I get the following by grepping `"Time for cloverdetratio_heatbath"` ad piping to `grep cloverdetratio2light`:

**yes_HB**: using the HB solver for `cloverdetratio2light`:

```
# : Time for cloverdetratio_heatbath 1.379339e+00 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.339430e+00 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.339117e+00 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.339778e+00 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.341887e+00 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
```

**no_HB**: using the HB solver for `cloverdetratio2light`:

```
# : Time for cloverdetratio_heatbath 4.108312e+01 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.089902e+01 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 9.869443e+00 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.000580e+01 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
# : Time for cloverdetratio_heatbath 1.104594e+01 s level: 1 proc_id: 0 /HMC/cloverdetratio2light:cloverdetratio_heatbath
```

The 1st is trajectory HB takes `~40s` due to the MG setup time. Since this has to be done anyway at some point, using the last 4 trajectories one finds that the `HB_solver` with `CG` uses `~13%` less time in the heatbath.

kostrzewa:
The problem with the github action was due to the github runner moving to Ubuntu 22.04 and there, the MPI libs are distributed separately from the compiler wrapper and utilities.

There's still another problem with the DDalphaAMG build because DDalphaAMG has issues with GCC 11.3 and how it organizes linking. I don't have the time to fix that problem right now.

I'll try to check the documentation tomorrow (not sure if I'll manage) and will pull this in then.

kostrzewa:
I've improved the github actions through "artifact upload" (https://docs.github.com/en/actions/using-workflows/storing-workflow-data-as-artifacts) which makes the various config.log (and output.data) accessible through the web interface post-build.

kostrzewa:
alright, finally fixed our CI ...

kostrzewa:
Damn, there's an issue with the initialisation of `HB_maxiter` for the "default" case (no HB solver defined).

With 

```
BeginMonomial CLOVERDETRATIO
  Timescale = 2
  kappa =    0.1400086
  2KappaMu = 0.000215613244
  rho = 0.0
  rho2 = 0.0015
  CSW = 1.7112
  AcceptancePrecision =  1.e-20
  ForcePrecision = 1.e-18
  Name = cloverdetratio3light
  solver = cg
  MaxSolverIterations = 100000
  UseExternalInverter = quda
  UseSloppyPrecision = single
EndMonomial
```

which should thus allow for 100k iterations in the HB, acceptance and derivative steps, I get:

```
# TM_QUDA: Using single prec. as sloppy!
# TM_QUDA: Called _loadGaugeQuda for gauge_id: 0.000000
# TM_QUDA: Theta boundary conditions will be applied to gauge field
# TM_QUDA: Time for reorder_gauge_toQuda 1.359407e-02 s level: 4 proc_id: 0 /HMC/cloverdetratio3light:cloverdetratio_heatbath/solve_degenerate/invert_eo_degenerate_quda/reorder_gauge_toQuda
# TM_QUDA: Time for loadGaugeQuda 2.442663e-01 s level: 4 proc_id: 0 /HMC/cloverdetratio3light:cloverdetratio_heatbath/solve_degenerate/invert_eo_degenerate_quda/loadGaugeQuda
# TM_QUDA: Using mixed precision CG!
# TM_QUDA: Using EO preconditioning!
# TM_QUDA: Time for loadCloverQuda 1.201410e-01 s level: 4 proc_id: 0 /HMC/cloverdetratio3light:cloverdetratio_heatbath/solve_degenerate/invert_eo_degenerate_quda/loadCloverQuda
# TM_QUDA: mu = 0.000770000000, kappa = 0.140008600000, csw = 1.711200000000
# TM_QUDA: Time for reorder_spinor_eo_toQuda 4.586891e-03 s level: 4 proc_id: 0 /HMC/cloverdetratio3light:cloverdetratio_heatbath/solve_degenerate/invert_eo_degenerate_quda/reorder_spinor_eo_toQuda
# QUDA: WARNING: Exceeded maximum iterations 5000
# QUDA: CG: Convergence at 5000 iterations, L2 relative residual: iterated = 2.350305e-09, true = 2.342661e-09 (requested = 1.000000e-10)
```
```

kostrzewa:
@simone-romiti can you find the logic bug? It seems to me that it should work correctly:

1. during `read_input`, `add_monomial` is called and sets the default params
2. in the remaining `read_input` for this monomial, `MaxSolverIterations` (in other words `mnl->maxiter`) is set
3. during `init_monomials`, for `DETRATIO` and `CLOVERDETRATIO`, the parameters from the "default" solver are taken over for the HB_solver. In particular, this should include `mnl->HB_maxiter = mnl->maxiter`

kostrzewa:
Thanks for checking, I agree with your conclusions. When I reviewed `monomial.c`, I missed that `no_monomials` was used instead of `i` for setting `HB_maxiter`.

I also agree that `DETRATMONOMIAL` should be added to line `2403` in `read_input.l`.

PR #547, opened 2022-08-23
Documentation gradient gflow

simone-romiti:
Following what is done in `read_input.l`, I've added the documentation for the gradient flow online measurement.


simone-romiti:
ok, thank you for the review and merge

PR #554, opened 2023-01-19
fix compiling error

Marcogarofalo:


quda_interface.c
```diff
@@ -1779,7 +1779,7 @@ void _setTwoFlavourSolverParam(const double kappa, const double c_sw, const doub
   }
 }
 
-_setMGInvertParam(QudaInvertParam * mg_inv_param, const QudaInvertParam * const inv_param){
+void _setMGInvertParam(QudaInvertParam * mg_inv_param, const QudaInvertParam * const inv_param){
```
kostrzewa, line 1782: wow, amazing that this worked anyway ;)

Marcogarofalo:
Is the merging blocked if the checks fail?

kostrzewa:
In principle, yes, but I know that the checks fail because the Ubuntu version on the github runner changed. I've just compiled it on QBIG without problems and will merge now.

PR #560, opened 2023-03-24
added interface for eigensolver on QUDA

aniketsen:


phmc.c
```diff
@@ -222,12 +226,31 @@ void phmc_compute_ev(const int trajectory_counter,
...
+                 mnl->solver_params.type, mnl->solver_params.rel_prec, 1, // we only support even-odd here
+                 mnl->solver_params.refinement_precision, mnl->solver_params.sloppy_precision, mnl->solver_params.compression_type);
 
+#endif
+
+  no_eigenvalues = 1;
   temp = eigenvalues_bi(&no_eigenvalues, max_iter_ev, eigenvalue_precision, 0, Qsq);
+
+#ifdef TM_USE_QUDA
+  if(mnl->EVMax == 1.) {
```
kostrzewa, line 241: this rescaling should be done in the QUDA interface

phmc.c
```diff
@@ -222,12 +226,31 @@ void phmc_compute_ev(const int trajectory_counter,
     printf("# Computing eigenvalues for heavy doublet\n");
   }
 
-  no_eigenvalues = 1;
+#ifdef TM_USE_QUDA
```
kostrzewa, line 229: I would like to avoid callling QUDA initialisation functions outside of `quda_interface.c` if possible.

phmc.c
```diff
@@ -222,12 +226,31 @@ void phmc_compute_ev(const int trajectory_counter,
...
+  if(mnl->EVMax == 1.) {
+    temp = temp / mnl->StildeMax;
+  }
+#endif
   
   no_eigenvalues = 1;
   temp2 = eigenvalues_bi(&no_eigenvalues, max_iter_ev, eigenvalue_precision, 1, Qsq);
+
+#ifdef TM_USE_QUDA
+  if(mnl->EVMax == 1.) {
```
kostrzewa, line 250: this rescaling should be done in the QUDA interface

quda_interface.c
```diff
@@ -2573,3 +2618,90 @@ void  compute_WFlow_quda(const double eps, const double tmax, const int traj, FI
...
+
+Interface function for Eigensolver on Quda
+
+*********************************************************/
+
+
+void eigsolveQuda(int n, _Complex double *host_evals, double tol, 
+                  int blksize, int blkwise,
+                  int max_iterations,  int maxmin) {
+
```
kostrzewa, line 2635: The philosophy in the other QUDA solver wrappers was always that they call any initialisation functions that they might require, rather than relying on QUDA being correctly initialised externally. It would be great if the same could be done here. Of course, this might require passing through parameters from the monomial, for example.

solver/eigenvalues_bi.c
```diff
@@ -132,6 +140,20 @@ double eigenvalues_bi(int * nr_of_eigenvalues,
     printf(" Values of   mu = %e     mubar = %e     eps = %e     precision = %e  \n \n", g_mu, g_mubar, g_epsbar, precision);
   }
 
+  /* For now, using the TM_USE_QUDA flag
+   * Ideally, one would use an operator flag
+   * like useExternalEigSolver. */
```
kostrzewa, line 145: I agree. Please add a flag to the ND*RAT monomials, something like `UseExternalEigSolver` as you suggest here to trigger the usage of QUDA for the eigensolver call.

quda_interface.c
```diff
@@ -2573,3 +2618,90 @@ void  compute_WFlow_quda(const double eps, const double tmax, const int traj, FI
...
+   * necessary. But have not found a way to consistently set
+   * the different precisions. */
+  eig_param.invert_param->cuda_prec_eigensolver = inv_param.cuda_prec;
+  eig_param.invert_param->clover_cuda_prec_eigensolver = inv_param.cuda_prec;
+
+  strncpy(eig_param.vec_outfile,"",256);
+  strncpy(eig_param.vec_infile,"",256);
+  
+
+  eig_param.n_conv = n;
```
kostrzewa, line 2699: do these defaults need to be commented on and/or exposed to be user-configurable?

kostrzewa (changes requested):
@aniketsen This is great, thanks a lot. There are a few things that need to be added / modified before this can be merged. It will also require a line or two in the documentation.

quda_interface.c
kostrzewa: let's just test for `quda_initialized = 1` here and `fatal_error` if it is not

quda_interface.c
```diff
@@ -2588,17 +2588,42 @@ Interface function for Eigensolver on Quda
...
+                    SloppyPrecision sloppy_precision, CompressionType compression) {
+
+  tm_stopwatch_push(&g_timers, __func__, "");
   
   // check if QUDA is initialized
   if (!quda_initialized) {
-    fatal_error("QUDA must be initialized.","eigsolveQuda");
-    return -1;
+    // it returns if quda is already init
+    _initQuda();
```
Marcogarofalo, line 2601: I think that here you can just call `_initQuda`. The if guard `if (!quda_initialized)` is redundant since it is checked inside `_initQuda`.

phmc.c
```diff
@@ -223,11 +228,36 @@ void phmc_compute_ev(const int trajectory_counter,
...
-
-  temp = eigenvalues_bi(&no_eigenvalues, max_iter_ev, eigenvalue_precision, 0, Qsq);
+  if(mnl->external_eigsolver == QUDA_EIGSOLVER) {
+    temp = eigsolveQuda(no_eigenvalues, eigenvalue_precision, 1, 0, max_iter_ev, 0,
+                        mnl->accprec, mnl->maxiter, mnl->solver, g_relative_precision_flag,
+                                                       1, // we only support even-odd here
+                                                       mnl->solver_params.refinement_precision,
+                                                       mnl->solver_params.sloppy_precision,
+                                                       mnl->solver_params.compression_type);
+    if( fabs(mnl->EVMax - 1) < 2*DBL_EPSILON ) {
```
kostrzewa, line 238: @aniketsen  Is this the right way around?

kostrzewa:
@aniketsen Can you change the source of this pull-request to the quda_work_eigensolver branch of etmc/tmLQCD (which I just created based on aniketsen:quda_work_eigensolver)? That way we can all push into it for minor tweaks and changes. (I wanted to add some documentation, for example)

kostrzewa:
I've now also merged with the latest `quda_work` there as `aniketset:quda_work_eigensolver` was not kept up to date with `etmc:quda_work` and was missing a number of commits (several of which related to the continuous integration stuff in .github)

phmc.c
```diff
@@ -223,11 +228,36 @@ void phmc_compute_ev(const int trajectory_counter,
   }
 
   no_eigenvalues = 1;
-
-  temp = eigenvalues_bi(&no_eigenvalues, max_iter_ev, eigenvalue_precision, 0, Qsq);
+  if(mnl->external_eigsolver == QUDA_EIGSOLVER) {
+    temp = eigsolveQuda(no_eigenvalues, eigenvalue_precision, 1, 0, max_iter_ev, 0,
```
kostrzewa, line 232: there needs to be a check here for `#ifdef TM_USE_QUDA` as the code must also compile without QUDA

In the `#else` branch of this `#ifdef`, there should be a failure with an informative message like "Attempted to use QUDA eigensolver but tmLQCD has been compiled without QUDA." (or something to that effect)

PR #564, opened 2023-05-24
first stab at using polynomial acceleration for the ND eigensolver

kostrzewa:


kostrzewa:
@aniketsen did you have a chance to play around with this?

aniketsen (approved):

PR #568, opened 2023-06-17
expose settings for the QUDA eigensolver

kostrzewa:


kostrzewa:
Just to let you know I've started exposing the settings for the QUDA eigensolver as I'm prepping to document it.

aniketsen:
I have also done this already,  but did not push it yet. But I was adding the parameters to the monomial, because I thought that these are dependent on the operator. But maybe I am wrong.

aniketsen:
And also I am passing the parameters as an input to the interface function, rather than having a global quda_input parameter. This would make it easier for me to write a tuning script.

aniketsen:
please take a look at #569 for my implementation.

kostrzewa:
> I have also done this already, but did not push it yet. But I was adding the parameters to the monomial, because I thought that these are dependent on the operator. But maybe I am wrong.

They are indeed dependent on the operator and for the completely general case I agree with you, of course. In that case, however, you will also need to add them to to the 'operator' struct, if at some point you want to measure the eigenvalues of some set of operators defined in the input file. I understand why you're focusing on the monomial now, of course.

> And also I am passing the parameters as an input to the interface function, rather than having a global quda_input parameter. This would make it easier for me to write a tuning script.

Sure, this would have been the next step.

> please take a look at https://github.com/etmc/tmLQCD/pull/569 for my implementation.

Let's go with that.

PR #571, opened 2023-10-21
allow writing out of flowed gauge field at specified intervals

kostrzewa:


kostrzewa:
I will add another component to this which allows the user to specify a list of flow times at which the field should be written to disk (rather than a single fixed interval).

PR #579, opened 2024-01-30
output the values of mubar, epsbar, kappa and csw passed to invert_doublet_eu_quda

kostrzewa:
FYI @aniketsen


kostrzewa:
I've tested that it compiles and works correctly already.

PR #582, opened 2024-02-06
added interface for saving gauge back to host

aniketsen:
added interface function `_saveGaugeQuda` to save gauge from QUDA back to host. Reordering back to tmLQCD format is performed with `reorder_gauge_fromQuda`.

### DIFF
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -584,6 +584,86 @@ void reorder_gauge_toQuda( const su3 ** const gaugefield, const CompressionType 
   tm_stopwatch_pop(&g_timers, 0, 0, "TM_QUDA");
 }
 
+void reorder_gauge_fromQuda( const su3 ** const gaugefield, const CompressionType compression ) {
+  tm_stopwatch_push(&g_timers, __func__, "");
+
+#ifdef TM_USE_OMP
+#pragma omp parallel
+  {
+#endif
+  _Complex double tmpcplx;
+
+  size_t gSize = (gauge_param.cpu_prec == QUDA_DOUBLE_PRECISION) ? sizeof(double) : sizeof(float);
+  
+  // now copy and reorder
+#ifdef TM_USE_OMP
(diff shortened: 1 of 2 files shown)


aniketsen:
@kostrzewa since `_saveGaugeQuda` is implemented now, I think `_loadGaugeQuda` also needs to be generalized to load any given gauge and not just `g_gauge_field`. Otherwise one can only save the gauge on device back to only `g_gauge_field` in order to load it back again. Also one needs a flag for reloading already loaded gauge back to device, since currently the only way to do it is to set `g_gauge_state.loaded` to `0` (which I did for testing, but removed in the commit, since it is not correct).

quda_interface.c
```diff
@@ -584,6 +584,86 @@ void reorder_gauge_toQuda( const su3 ** const gaugefield, const CompressionType
...
+          int tm_idx = x3 + LZ*x2 + LY*LZ*x1 + LX*LY*LZ*x0;
+#endif
+          int oddBit = (x0+x1+x2+x3) & 1;
+          int quda_idx = 18*(oddBit*VOLUME/2+j/2);
+
+          if( compression == NO_COMPRESSION && quda_input.fermionbc == TM_QUDA_THETABC ) {
+            // apply theta boundary conditions if compression is not used
+            for( int i=0; i<9; i++ ) {
+              tmpcplx = gauge_quda[0][quda_idx+2*i] + I*gauge_quda[0][quda_idx+2*i+1];
+              tmpcplx *= -g_kappa/phase_1;
```
kostrzewa, line 620: @aniketsen I think I don't quite understand how this removes the TBC phases.

kostrzewa: I somehow didn't see that the ratio was inverted... too tired I guess.

kostrzewa:
We should still discuss how to make sure that `gauge_quda` is reset when it is next used as in the current situation `g_gauge_field` and `gauge_quda` will of course go out of sync...

PR #610, opened 2024-03-15
WIP eigenvalues online measurement which should hopefully work for all supported operators

kostrzewa:
need to see how to unify the CPU and GPU versions, GPU should work already


kostrzewa:
For now this is just FYI

PR #611, opened 2024-03-16
suppress message about MgSetup2KappaMu if it does not differ from current mu

kostrzewa:
I've been wanting to fix this for ages now.


kostrzewa:
works as expected

PR #613, opened 2024-04-24
in the hmc_mk2 profiler, optionally include all timings

kostrzewa:
Through a third command line argument, in the per-monomial tables and plots rather than grouping timings below 5% in the 'other' category, all timings can now be shown.


kostrzewa:
image

Marcogarofalo:
could we set also the digits to 3 in the tables? there should be enough space and will be useful when looking at small things.

kostrzewa:
> could we set also the digits to 3 in the tables? there should be enough space and will be useful when looking at small things.

Done.

image

Marcogarofalo:
Thank you!

Marcogarofalo (approved):

PR #617, opened 2025-02-06
add  mg_param->n_vec_batch[level]

Marcogarofalo:
Do we also need some compile time variable to compile with an old version of QUDA? 
Something like `#ifdef TM_QUDA_mrhs`.

### DIFF
--- a/.github/workflows/basic-build.yaml
+++ b/.github/workflows/basic-build.yaml
@@ -23,14 +23,14 @@ jobs:
         run: echo "Number of cores ${nb_cores}"
 
       - name: get_packages
-        run: sudo apt-get update && sudo apt install openmpi-bin openmpi-common libopenmpi3 libopenmpi-dev numdiff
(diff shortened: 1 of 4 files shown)


kostrzewa:
Hi. Thanks for this. I think wrapping this in `TM_QUDA_EXPERIMENTAL` is sufficient as we don't have any other things which really need to still be wrapped. This also means that basically all of the existing `#if[n]def TM_QUDA_EXPERIMENTAL` can be removed, I think.

kostrzewa:
Thanks a lot also for fixing the tests!!

kostrzewa:
Changes don't seem to affect the HMC. Tested with new and old QUDA versions.

I have confirmed in the meantime that the current QUDA develop on daint.alps (sometimes) leads to non-deterministic NaNs in the clover term for some reason. Instead, the HMC seems to run fine over "many trajectories" with an older QUDA commit.

PR #620, opened 2025-02-17
Some fixes required to build TMLQCD on Daint with libsmear

gkanwar:
This PR includes a fix for one compilation issue and turns on the libsmear library with fixes for the various headers to ensure that it builds correctly. If libsmear is intentionally disabled by default, I can also redo the PR removing those bits (i.e. just leaving the one fix for the compilation bug).

PR #621, opened 2025-02-17
Some fixes required to build TMLQCD on Daint with libsmear

gkanwar:
This PR includes a fix for one compilation issue and turns on the libsmear library with fixes for the various headers to ensure that it builds correctly. This comments out the new(?) `gauge_field_t` interfaces so the library can be build using the old(?) `su3_gauge**` tuple signatures.

See also PR #620

### DIFF
--- a/configure.in
+++ b/configure.in
@@ -42,7 +42,7 @@ LDFLAGS="$LDFLAGS -L\${HOME}/lib -L\${top_builddir}/lib"
 CCLD=${CC}
 
 # compilation in operator is slowest so we do it first, saves time in parallel compiles
-USESUBDIRS="operator linalg solver monomial buffers cu io meas xchange init rational wrapper"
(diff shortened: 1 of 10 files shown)


gkanwar:
Hi @kostrzewa , I think this PR should now be in a mergeable state. Let me know if you want any further edits!

kostrzewa:
Have to check compilation / running manually on a few systems for changes which touch the linking. I've found some issues on one machine but I'm not sure if they relate to your changes or simply changes on the machine or previous changes in tmLQCD...

Did you run the HMC with this by any chance on some pure CPU machine with QPhiX ?

kostrzewa:
From what I can see the fact that smearing is built should not actually affect the linking. I think the issue that I'm seeing might be limited to the machine that I'm testing on. Let's merge this and fix it if it breaks somewhere.

kostrzewa (approved):

PR #623, opened 2025-02-28
only use MGSetup2KappaMu if it has actually been set

### LINKED ISSUE #622: need more logic for dealing with mg_setup_2kappamu
Right now, if `MGSetup2KappaMu` is not set, the default value is `0.0` and the MG setup will be performed with `mu=0.0`.

We should check for `fabs(quda_input.mg_setup_2kappamu) > 2*DBL_EPSILON` in addition to checking for the difference to the inversion mu.

kostrzewa:
In principle this should fix #622, test run is queued on Juwels Booster

### DIFF
--- a/quda_interface.c
+++ b/quda_interface.c
@@ -1604,7 +1604,8 @@ void _updateQudaMultigridPreconditioner(){
 
     // if we have set an explicit mu value for the generation of our MG setup,
     // we would like to use it here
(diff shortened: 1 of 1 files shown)


kostrzewa:
@NKalntis @fernandezdlg  @uwenger @gkanwar This fixes the issue that you were seeing when `MGSetup2KappaMu` was not specified.

PR #646, opened 2025-10-27
Formatting

mtaillefumier:
Apply formatting to the entire code base. This PR also add the `su3.h` file to headers files requiring it. This issue was not found before because the header ordering was not alphabetic.

PR #648, opened 2025-10-29
fix two type errors for scalar builds

urbach:
this should fix scalar builds


mtaillefumier:
This PR is merged in PR #646  and the commit is first on the commit list.

PR #654, opened 2025-11-26
Cleanup/qphix interface

mtaillefumier:
- Moved the qhpix interface to a dedicated directory
- Updated the build system


mtaillefumier:
I agree with you. I am still puzzled by this error and autotools can be very obscure some times. 

I would suggest I close the PR for the time being.

PR #662, opened 2026-01-10
Fix implicit declaration compiling with qphix

Finkenrath:
Fix compiler warning for implicit function declaration, while compiling tmLQCD with QPhiX.
- add the missing headers


mtaillefumier (approved):
LGTM

PR #668, opened 2026-03-06
[WIP DO NOT MERGE] switch kostrzewa/ddalphaamg to etmc/ddalphaamg

kostrzewa:
Let's see what this does to the tests.


kostrzewa:
> If we compile it in the exact same way, tests should agree without changing solver precisions, no?

I'm not sure how and with which commit of DDalphaAMG I generated the reference output at the time so I'm not too surprised that we see deviations.

I agree that without threading we should obtain agreement again if we increase the solver precision throughout and regenerate the reference data (which should ideally be done on a different machine).

PR #671, opened 2026-03-24
use stringent solver precisions for ddalphaamg CI workflow

kostrzewa:
Let's see if increasing solver precisions fixes our DDalphaAMG reproducibilty problem.

I've also removed two orphaned symlinks related to the CSCS CI which became orphaned when that was restructured.


chaoos (approved):
Looks good to me, thanks for removing the broken links.

PR #672, opened 2026-03-27
include hopping.h and hoping_body_dbl.inc in tm_[sub,times]_Hopping_Matrix.c 

kostrzewa:


mtaillefumier (approved):
i will merge it immediately

PR #673, opened 2026-03-27
Konstantin devel Wilson+Clover N_f=2+1 implementation

kottnad:
Provides a working implementation of N_f=2+1 (plain) Wilson+Clover simulations.
Still lacks a GPU implementation of the force for the single flavor part though.


kostrzewa:
@chaoos @mtaillefumier any thoughts on the best order of merging this or #664 ?

Issue #684, opened 2026-04-17
DetectSimdAndAlignment detects AVX512 on AMD EPYC (znver2)


kostrzewa:
```
-- Performing Test _HAVE_AVX512_DOUBLE
-- Performing Test _HAVE_AVX512_DOUBLE - Success
-- SIMD: AVX512 (x86), align=64
```

kostrzewa:
Should the autodetection on Grace use a larger alignment? This is what it finds right now:

```
-- Performing Test _HAVE_SVE
-- Performing Test _HAVE_SVE - Success
-- SIMD: SVE (ARM), align=16
```

mtaillefumier:
This issue should be fixed by PR#689. The root cause was that compilation does not imply that a given CPU supports the instructions. To solve the problem, we must compile the test code and run it locally.

mtaillefumier:
this issue is resolved by PR#689

PR #689, opened 2026-04-20
Cmake fix

mtaillefumier:
- The SIMD tests are now compiled and executed (issue #684 )
- Naming convention for the cmake config files is uniform (issue #691)
- Updated the commit version of lemon (PR #690)
- Solve the installation directories inconsistencies  (issue #691)


mtaillefumier:
@chaoos : I would appreciate your feedback on this PR. It needs to be merged soonish as it fixes few issues with include directories.

kostrzewa:
@kottnad could you please review this and merge?

chaoos:
@mtaillefumier  Looks good to me.

Only small concern I have is that you changed multiple `check_cxx_source_compiles` into `check_source_runs`. That might lead to problems when not building on a target node. This is no problem on daint, but maybe on another cluster. Is there a possibility to bypass the SIMD detection and provide values explicitely?

mtaillefumier:
True. I just do not know how to check which instruction is effectively supported during compile time without running it after compilation.  `check_cxx_source_compiles` only checks that the code compiles not that the instruction is supported on the machine at runtime, that's why my initial implementation was defective in practice. 

The best solution would be to check if the instruction set is supported at compilation time without running the executable but I am not sure if it is possible in practice. 

Is there any case of cross-compilation ?

chaoos:
I understand, please go on. We can think about adding a bypass when the problem arises. For me this can be merged, thanks for the changes! 👍

kottnad (approved):

PR #697, opened 2026-06-12
Added all compiling tests from the test directory

mtaillefumier:
- Added all tests from the tests directory that compile
- Removed Makefile.in
- Remove variable conflict between tmlqcd and one of the tests

PR #698, opened 2026-06-16
use latest lemon commit so that the tmLQCD install step works

kostrzewa:


kostrzewa:
cscs-ci run default

kostrzewa:
@mtaillefumier If the CSCS-CI run will fail on #697 too I would very much appreciate your help in figuring out what's going on.

mtaillefumier:
Yeah. there are some issues with spack itself. i will have a look.

mtaillefumier:
cscs-ci run default

mtaillefumier:
cscs-ci run default

mtaillefumier:
cscs-ci run default

mtaillefumier:
cscs-ci run default

mtaillefumier:
cscs-ci run default

mtaillefumier:
The CI/CD is working again. I had to add the c-lime package to the local `spack ` packages because linking was failing.

kostrzewa:
Thanks! Yes, locally we had to add `-fPIC` to the `c-lime` builds as well for linking to work.

I will need to check again but frankly I think we can compile and use tmLQCD completely without lime if we're using lemon.

kostrzewa:
@kottnad since I can't approve my own PR I would be grateful if you could take a look at this at some point. Thanks!

kottnad (approved):

PR #705, opened 2026-08-31
fix broken parsing of DETRATIO monomial

### LINKED ISSUE #704: Init of DETRATIO is broken
### DETRATIO monomials cannot be parsed: `DETRATMONOMIAL` start condition missing from most rule blocks

(bug found and issue description generated with help from Claude Opus 5)

A minimal `DETRATIO` monomial fails on its very first parameter line:

```
BeginMonomial DETRATIO
  Timescale = 2
  2KappaMu = 0.015
  2KappaMu2 = 0.6
  kappa = 0.15
  kappa2 = 0.15
  AcceptancePrecision = 1e-24
  ForcePrecision = 1e-24
  Name = detratio1
  Solver = CG
EndMonomial
```

```
initialising monomial with type DETRATIO 1 line 39
monomial has id 2
Parsing error in line 40
Aborting...!
Could not make sense out off:  Timescale = 2
```

#### Cause

`src/lib/read_input.l:2285` dispatches `DETRATIO` into the exclusive start condition
`DETRATMONOMIAL`, but that state is missing from the newline rule at
`src/lib/read_input.l:3732`, which is what consumes an end-of-line *without* leaving
the current block. The newline therefore falls through to the catch-all at
`src/lib/read_input.l:3735`, which does `BEGIN(0)`. The scanner is back in `INITIAL`
before the first parameter line is ever read, so `Timescale` (or whatever else) hits `<*>.`
(`src/lib/read_input.l:3740`) and then the error rule at
`src/lib/read_input.l:3743`.

Introduced in 2534c6c ("HB_solver specified only for the cloverdetratio and detratio
monomials", 2022-07-12), which added `else if(mnl->type == DETRATIO) BEGIN(DETRATMONOMIAL);`
so that `HB_Solver` could be set separately for `DETRATIO`, and added the new state only to
the `Timescale` and `HB_*` blocks. Before that commit `DETRATIO` fell through to
`else BEGIN(DETMONOMIAL);`, and `DETMONOMIAL` carries the complete keyword set a detratio
monomial needs.

#### Fix

Add `DETRATMONOMIAL` to the start-condition lists at `src/lib/read_input.l` lines 2337 (kappa),
2345 (2kappamu2/kappa2), 2402 (ExternalInverter and corresponding parameters), 2558 (2kappamu and some other stuff) and 3732 (the monomial dispatch itself).

I will create a PR with the fix soon

aniketsen:
fixes #704


kostrzewa:
That's super weird. Support for the DETRATMONOMIAL must have disappeared by mistake at some point... a long time ago, way before the commit that Opus identified there.

kostrzewa:
In fact it's not just weird but completely crazy because we generated 2+1+1 non-clover ensembles to replace some broken ones...

kostrzewa (approved):

aniketsen:
to my understanding `DETRATIO` used the same parsing logic as `DET`

even going back as far as 890415c, I can not find a separate logic for DETRATIO. It always falls back to `else BEGIN(DETMONOMIAL);` block.

I think its only when 2534c6c added a separate block for DETRATIO, this bug arose, since it was not added everywhere that was necessary. I am a bit surprised though that this went completely unnoticed

kostrzewa:
Yeah, you're right. In fact the logic is now a bit broken because strictly speaking, `2KappaMu2` and `Kappa2` should only be possible for the ratio.

kostrzewa:
> I am a bit surprised though that this went completely unnoticed

I'm not, the non-clover action has not been used since 2014 or so and it's not explicitly included in any tests.

aniketsen:
> In fact the logic is now a bit broken because strictly speaking, 2KappaMu2 and Kappa2 should only be possible for the ratio.

yes, completely agree. But I suppose they are not used anywhere by the DET monomial. But I agree, should be cleaned up now.

PR #706, opened 2026-09-07
Removed the OpenMP option from the input files

mtaillefumier:
- OMP is controlled with the OMP_NUM_THREADS variable.
- Issue: QPhiX parameters depend on the the number of threads.
- Still need to clear all examples


src/lib/read_input.l
```diff
@@ -932,11 +932,6 @@ static inline double fltlist_next_token(int * const list_end){
     mg_update_setup_iter=a;
     if(myverbose) printf("  MG_Update_Setup_Iter set to %d line %d operator %d\n", mg_update_setup_iter, line_of_file, current_operator);
   }
-  {SPC}*MGOMPNumThreads{EQL}{DIGIT}+ {
-    sscanf(yytext, " %[a-zA-Z] = %d", name, &a);
-    mg_omp_num_threads=a;
```
kostrzewa, line 937: @mtaillefumier the logic in DDalpaAMG_interface.c needs to be adjusted and the global `mg_omp_num_threads` variable should be removed.

doc/sample-input/sample-dfl.input
```diff
@@ -26,7 +26,7 @@ SourceType = timeslice
 #ReadSource = no
 #NoSamples = 1
 UseSloppyPrecision = no
-OmpNumThreads = 2
```
kostrzewa, line 29: This should actually be deleted.

.github/workflows/nf211-qphix-tmclover.yaml
```diff
@@ -192,7 +192,7 @@ jobs:
       - name: nf2_qphix_hmc_tmcloverdetratio
         working-directory: ${{github.workspace}}/main/build
         run: |
-          mpirun -np 2 src/bin/hmc_tm \
```
kostrzewa, line 198: this will conflict with current master now since the test workflows have been replaced by more complete ones. Can you please make sure that OMP_NUM_THREADS is set correctly for all of them?

kostrzewa:
> Issue: QPhiX parameters depend on the the number of threads.

I don't think there's any problem. `Ncores`, `Sy` and `Sz` need to be set correctly by the user and their product is compared with `omp_num_threads`, which is now set via `OMP_NUM_THREADS` via `init_openmp`.

